PR #3, opened 2025-01-20
Vasp mapping parser

ladinesa:


src/nomad_simulation_parsers/parsers/utils.py
JFRudzinski, line 1: I think we will need a utils folder, e.g., for the MDAnalysis parser and MDParser

JFRudzinski (approved):
Perhaps @ndaelman-hu also wants to check this

src/nomad_simulation_parsers/parsers/__init__.py
```diff
@@ -15,9 +23,27 @@ def load(self):
...
     mainfile_contents_re=r'EXCITING.*started[\s\S]+?All units are atomic ',
     mainfile_name_re=r'^.*.OUT(\.[^/]*)?$',
     code_name='exciting',
     code_homepage='http://exciting-code.org/',
 )
+
+vasp_parser_entry_point = EntryPoint(
+    name='parsers/vasp',
+    description='Parser for VASP XML and OUTCAR outputs',
+    parser_class_name='nomad_simulation_parsers.parsers.exciting.parser.VASPParser',
```
ndaelman-hu: Why is the `VASPParser` listed under `exciting`?

src/nomad_simulation_parsers/parsers/__init__.py
```diff
@@ -15,9 +23,27 @@ def load(self):
...
+    python_package='nomad_simulation_parsers',
+    code_name='VASP',
+    mainfile_contents_re=(
+        r'^\s*<\?xml version="1\.0" encoding="ISO-8859-1"\?>\s*?\s*<modeling>?\s*'
+        r'<generator>?\s*<i name="program" type="string">\s*vasp\s*</i>?|'
+        r'^\svasp[\.\d]+.+?(?:\(build|complex)[\s\S]+?executed on'
+    ),
```
ndaelman-hu, line 44: I would put the HDF5 parser above xml in the matching priority

src/nomad_simulation_parsers/parsers/__init__.py
```diff
@@ -15,9 +23,27 @@ def load(self):
...
+    mainfile_mime_re='(application/.*)|(text/.*)',
+    mainfile_name_re='.*[^/]*xml[^/]*',
+    mainfile_alternative=True,
+    supported_compressions=['gz', 'bz2', 'xz'],
```
ndaelman-hu, line 48: Shouldn't the compression support be specified at a higher level?

src/nomad_simulation_parsers/parsers/exciting/parser.py
ndaelman-hu, line 1: I like this approach, very tidy. This `close` function is new, right?

src/nomad_simulation_parsers/parsers/utils.py
ndaelman-hu, line 1: Please explain why annotations have to be removed in the pydocs.
From the code below, I take that it removes the old key. But could one simply not call that key, correct?

src/nomad_simulation_parsers/schema_packages/vasp.py
ndaelman-hu, line 1: This is mapping to the old schema, correct?
You can merge this, but I'll have to redo it for the new schema

src/nomad_simulation_parsers/parsers/vasp/parser.py
```diff
@@ -0,0 +1,38 @@
...
+
+        # import schema to load annotations
+        from nomad_simulation_parsers.schema_packages import vasp
+
+        if 'outcar' in mainfile.lower():
+            parser = VASPOutcarParser()
+        else:
+            parser = VASPXMLParser()
+        # TODO remove this for debug
+        self.parser = parser
```
ndaelman-hu: what's the purpose of storing the `parser` here?

src/nomad_simulation_parsers/parsers/vasp/xml_parser.py
```diff
@@ -0,0 +1,77 @@
...
+            return
+
+        parser = Path(path=path)
+        return parser.get_data(source)
+
+    def get_forces(self, source: dict[str, Any]) -> dict[str, Any]:
+        value = self.get_data(source, path='.varray.v')
+        if value is None:
+            return {}
+        return dict(forces=value, npoints=len(value), rank=[3])
```
ndaelman-hu, line 50: Why are we still using this older style for extracting forces? Can't we use the mapping annotation here too?

src/nomad_simulation_parsers/parsers/utils.py
ladinesa: If we do not add then remove the added annotations every time we call parse, the annotations from the other parsers will be picked up and may override our intended annotation if the same key is used.  added doc

ladinesa: yes, i will convert it to a folder

src/nomad_simulation_parsers/parsers/__init__.py
ladinesa: I think it is in the correct place, in MatchingParser

src/nomad_simulation_parsers/parsers/__init__.py
ladinesa: which hdf5 parser?

src/nomad_simulation_parsers/parsers/__init__.py
ladinesa: ah yes my bad

src/nomad_simulation_parsers/parsers/vasp/parser.py
ladinesa: see TODO, will remove

src/nomad_simulation_parsers/schema_packages/vasp.py
ladinesa: i am still targeting this branch `100-moving-simulation-workflow-definitions-to-this-repo` please advise which is the latest so I can make the necessary changes

src/nomad_simulation_parsers/parsers/vasp/xml_parser.py
ladinesa: there is a bug for the def of forces i think so i have to define the rank directly here.

src/nomad_simulation_parsers/parsers/exciting/parser.py
ladinesa: i added the close context for a while now,

PR #30, opened 2025-08-08
Migrate nomad parser h5md

JFRudzinski:


JFRudzinski:
@ladinesa Could you check the parser and schema entry points here? Something is not connecting properly but I can't figure out what I missed, or what mistake I made in the transfer

ladinesa:
> @ladinesa Could you check the parser and schema entry points here? Something is not connecting properly but I can't figure out what I missed, or what mistake I made in the transfer

perhaps you did not disconnect the h5md parser from atomisticparsers? They have the same matching specs so if you run nomad parse locally it may match the one in atomisticparses

ladinesa:
there is also this one structlog error where a logger cannot be propagated to the child class and happens only in the pipeline. I only have a temporary fix for now see e.g. https://github.com/FAIRmat-NFDI/nomad-parser-plugins-simulation/blob/develop/src/nomad_simulation_parsers/parsers/abinit/parser.py#L448 . You have to do it for all your MappingParser classes.

src/nomad_simulation_parsers/parsers/h5md/parser.py
```diff
@@ -0,0 +1,328 @@
...
+                    step_data['value'] = step_data['value'].magnitude
+            custom_outputs.append({'name': key, **step_data})
+        return custom_outputs
+
+
+class H5MDParser(MDParser):
+    def __init__(self) -> None:
+        super().__init__()
+        self.h5_parser = H5MDH5Parser()
+        self.simulation_parser = MetainfoParser()
```
ladinesa: you also have to create a `MetainfoParser` class with the `logger` property

JFRudzinski:
Thanks @ladinesa I think I understand the issue now, I am trying to make a fix now

JFRudzinski:
@ladinesa 

I have solved the logger issue, but have been struggling to get the entry points to work properly. I now restructured my parser to mimic yours with an ArchiveWriter. However I am still having some fundamental issue where the test_schema_packages.py is failing 

I'd really appreciate it if you can have a look again and try to see what I am doing wrong

src/nomad_simulation_parsers/parsers/h5md/parser.py
```diff
@@ -0,0 +1,360 @@
...
+class H5MDArchiveWriter(ArchiveWriter):
+    def __init__(self):
+        self.h5_parser = H5MDH5Parser()
+        self.simulation_parser = H5MDMetainfoParser()
+        self.simulation_parser.max_nested_level = 10
+        self.workflow_parser = H5MDMetainfoParser()
+
+    def write_to_archive(self) -> None:
+        # reload schema annotations
+        reload(h5md)  # ? what exactly does this do?
```
ladinesa: this is necessary as all the annotations are written in the general schema, so when another parser is executed the h5md annotations would have been overwritten. top-level imports are cached so a brute force reload is necessary.

ladinesa:
> @ladinesa
> 
> I have solved the logger issue, but have been struggling to get the entry points to work properly. I now restructured my parser to mimic yours with an ArchiveWriter. However I am still having some fundamental issue where the test_schema_packages.py is failing
> 
> I'd really appreciate it if you can have a look again and try to see what I am doing wrong

pushed fix, you forgot to pass the parser kwargs to your init func, so h5md parser matched all files including the test file in test_schema_packages.

JFRudzinski:
@ladinesa Thank you!! :pray: :pray: 

These last 2 tests that are failing pass locally, so not sure how to address them. FYI - I made sure to check out the latest develop of nomad-lab locally.

For the h5md test error: If I remove this failing assertion line the rest of the test passes. Have you seen this behavior before of populating empty repeating subsections?

For the abinit error: I also check the pint version which is the same locally for me. Maybe you have some other ideas

ladinesa:
> @ladinesa Thank you!! 🙏 🙏
> 
> These last 2 tests that are failing pass locally, so not sure how to address them. FYI - I made sure to check out the latest develop of nomad-lab locally.
> 
> For the h5md test error: If I remove this failing assertion line the rest of the test passes. Have you seen this behavior before of populating empty repeating subsections?
> 
> For the abinit error: I also check the pint version which is the same locally for me. Maybe you have some other ideas

The h5md error is caused by the nested workflow issue that esma brought up last time and has since been fixed but not yet included in the current release. I suggest you comment it out for now until it gets picked up. The abinit fix is already in develop so i rebased your branch

src/nomad_simulation_parsers/parsers/__init__.py
```diff
@@ -106,6 +106,33 @@ class Wannier90ParserEntryPoint(EntryPoint):
...
+    description='NOMAD parser for H5MD.',
+    python_package='nomad_simulation_parsers',
+    mainfile_binary_header_re=b'^\\x89HDF',
+    mainfile_contents_dict={'__has_all_keys': ['h5md']},
+    mainfile_mime_re='(application/x-hdf)',
+    mainfile_name_re=r'^.*\.(h5|hdf5)$',
+    parser_class_name='nomad_simulation_parsers.parsers.h5md.parser.H5MDParser',
+    code_name='H5MD',
+    code_category='MD code',
+    # metadata={
```
ladinesa, line 121: remove this as no longer used?

src/nomad_simulation_parsers/parsers/h5md/parser.py
```diff
@@ -0,0 +1,358 @@
...
+from structlog.stdlib import BoundLogger
+
+from nomad_simulation_parsers.parsers.utils.general import remove_mapping_annotations
+from nomad_simulation_parsers.parsers.utils.mdparserutils import MDParser
+from nomad_simulation_parsers.schema_packages import h5md
+from nomad_simulation_parsers.schema_packages.h5md import Simulation
+
+LOGGER = get_logger(__name__)
+# Make base class non-abstract by providing the property globally
+if not hasattr(MetainfoParser, 'logger'):
```
ladinesa: I tried this but simply setting the attr does not not seem to work.

src/nomad_simulation_parsers/parsers/h5md/parser.py
```diff
@@ -0,0 +1,358 @@
...
+    # TODO temporary fix for structlog unable to propagate logger
+    @property
+    def logger(self):
+        return LOGGER
+
+    def get_value(self, name: str, dct: dict[str, Any]) -> Any:
+        value = dct.get(name, {})
+        if not isinstance(value, dict):
+            return value
+        value = value.get(self.value_key)
```
ladinesa, line 42: isn"t `value_key ` supposed to be an arg? If not maybe put self.value_key in class init

ladinesa: oh sorry my bad, this is hdf5 parser thing.

src/nomad_simulation_parsers/parsers/__init__.py
JFRudzinski: Oh yes, I wanted to ask you about that, is this section completely deprecated? Or if not is it useful? I was still defining some of these, but I just commented it out while I was trying to fix the errors just in case

ladinesa: This is not anymore used by the parser entry point. It was used in the old plugin mechanism.

src/nomad_simulation_parsers/parsers/h5md/parser.py
```diff
@@ -0,0 +1,358 @@
...
+            return {}
+
+        path_segments = path.split('.', 1)
+        source = parent.get(path_segments[0], {})
+
+        if len(path_segments) == 1:
+            return source
+        return self.get_source(source, path_segments[1])
+
+    def map_value(self, source: dict[str, Any], **kwargs) -> Any:
```
ladinesa: a little nitpicking but i think you can put key and enum_spec directly as args

ladinesa: you can actually make enum_spec as a function so you do not have the if in l74-77 also more general

ladinesa: or is this somehow used in the schema? then forget my previous comment

src/nomad_simulation_parsers/parsers/h5md/parser.py
```diff
@@ -0,0 +1,358 @@
...
+
+    def get_traj_data(self, source: dict[str, Any]) -> list[dict[str, Any]]:
+        positions = self.get_source(source, 'position')
+        steps = self.get_value('step', positions)
+        times = self.get_value('time', positions)
+        positions = self.get_value('value', positions)
+        velocities = self.get_source(source, 'velocity')
+        velocities = self.get_value('value', velocities)
+
+        assert len(steps) == len(times) == len(positions)
```
ladinesa: simply log the error here but return.

src/nomad_simulation_parsers/parsers/h5md/parser.py
```diff
@@ -0,0 +1,358 @@
...
+                return {}
+            times = self.get_value('time', dct)
+            if len(steps) != len(times):
+                self.logger.error(
+                    'Inconsistent step-time combinations in observable data.'
+                )
+
+            return {step: times[n] for n, step in enumerate(steps)}
+
+        def get_observable_steps(
```
ladinesa, line 206: i suggest you return the output dict here, so you do not redefine output_steps as a list later

src/nomad_simulation_parsers/parsers/h5md/parser.py
```diff
@@ -0,0 +1,358 @@
...
+        self.workflow_parser = H5MDMetainfoParser()
+
+    def write_to_archive(self) -> None:
+        # reload schema annotations
+        reload(h5md)
+
+        # create h5 parser
+        self.h5_parser.filepath = self.mainfile
+
+        self.trajectory_steps = Path(path='particles.all.position.step').get_data(
```
ladinesa, line 315: you do not have `self.trajectory_steps` in you init

src/nomad_simulation_parsers/parsers/h5md/parser.py
```diff
@@ -0,0 +1,358 @@
...
+            self.h5_parser.data, default=[]
+        )
+        self.h5_parser.trajectory_steps = self.trajectory_steps
+
+        # TODO consider using a single parser for the whole archive
```
ladinesa, line 320: I think you can do this if you start annotating from EntryArchive, I tried with fhiaims but I do not anymore remember why i did not pursue.

ladinesa (approved):

src/nomad_simulation_parsers/parsers/h5md/parser.py
JFRudzinski: Do you think I can remove the note then?

ladinesa: It can stay.

src/nomad_simulation_parsers/parsers/h5md/parser.py
JFRudzinski: Ah, this is a bit annoying about changing the structure...

I changed MDParser(Parser) to MDParser(MatchingParser), but now I guess I want H5MDArchiveWriter(MDParser, ArchiveWriter) & H5MDParser(MatchingParser)? Could this even work, or how should I deal with my MDParser() util class?

ladinesa: You can simply have your main parser be MatchingParser and your writer be MDParser(ArchiveWriter), sorry I really wanted to separate the 'parser' that interfaces with nomad, containing the infra-related stuff to the actual 'parser'  which does the actual writing to the archive.

src/nomad_simulation_parsers/parsers/h5md/parser.py
JFRudzinski: it's not in the schema, I am just trying to enforce the enum conventions within the schema, but allow people to have various capitalizations in the raw data. 

What do you mean as a function?

ladinesa: instead of enum_spec  as string, have something like enum_spec: lambda x: x.upper()

ladinesa: but i saw this in the schema in `molecular_dynamics.ThermostatParameters.thermostat_type`, oh sorry i misread your comment. anyway my suggestion is moot as you cannot provide functions in the annotations

JFRudzinski: ok, I just pass them now as args :sweat_smile:

JFRudzinski:
@ladinesa please note my last commit fixing errors induced from https://github.com/nomad-coe/nomad-simulations/pull/213 and https://github.com/nomad-coe/nomad-simulations/pull/214

Please let me know if they are fine and then I will merge this

ladinesa (approved):

PR #35, opened 2025-08-15
Implement parsing of vasp charge density

ladinesa:
Implementation of charge density parsing in vasp.


JFRudzinski:
@ladinesa switched myself out for @ndaelman-hu for review

JFRudzinski:
> @ladinesa switched myself out for @ndaelman-hu for review

Ok I got a bit confused here, I didn't realize this was in reference to the hdf5dataset issue. 

I guess I will leave us both as reviewers then...?

ladinesa:
@JFRudzinski I now put the definition for the charge density temporarily in the vasp schema. I would like to merge this as is as it becomes challenging to rebase. We just update this again to use the definition in nomad-simulations once available.

JFRudzinski (approved):

PR #37, opened 2025-08-18
LAMMPS parser: system parsing

### LINKED ISSUE #42: Parse system, excluding hierarchy. Update dependencies to new schema.

Bernadette-Mohr:
This PR addresses migrating `parse_system` of the LAMMPS parser from the old version in `atomisticparsers` to the new `ModelSystem`.

**TODO**
- [x] Update LAMMPS parser to new schema -- parse system only, without molecule hierarchy
- [x] Transfer tests from old parser, comment out non-system tests for now
- [x] Transfer Leo's changes to old parser - Mostly commented out and untested
- [x] Test parsing of CG systems -- should work using ParticleState()


pyproject.toml
```diff
@@ -33,7 +33,7 @@ maintainers = [
 license = { file = "LICENSE" }
 dependencies = [
     "nomad-lab>=1.3.16.dev100",
-    "nomad-simulations@git+https://github.com/nomad-coe/nomad-simulations.git@develop",
+    "nomad-simulations",  # @git+https://github.com/nomad-coe/nomad-simulations.git@develop",
```
JFRudzinski: you should not have to change these dependencies when you work in the distro dev, don't forget to change it back in the end

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1453 @@
...
+        ]
+
+        units = self.log_parser.units
+
+        def apply_unit(value, unit):
+            if not hasattr(value, 'units'):
+                value = value * units.get(unit, 1)
+            return value
+
+        def get_composition(children_names):
```
JFRudzinski: fyi - composition_formula will now be set by the Simulation normalizer in nomad-simulations, so you do not need to set this in the parser

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1453 @@
...
+        self.n_atoms = [self.traj_parsers.eval('get_n_atoms', n) for n in range(n_traj)]
+        self.trajectory_steps = [
+            step
+            for n in range(n_traj)
+            if (step := self.traj_parsers.eval('get_step', n)) is not None
+        ]
+
+        units = self.log_parser.units
+
+        def apply_unit(value, unit):
```
JFRudzinski: This probably should be a class method right? I imagine it will be used by other parsing functions

JFRudzinski (commented):
looking good so far, just a few minor comments upon browsing

pyproject.toml
Bernadette-Mohr: I unfortunately do have to do that, which doesn't seem to be too surprising: 
https://discord.com/channels/1201445470485106719/1383005977586172004

I'll make a habit of changing it back before I request a review.

Currently, I'm not seeing this behavior anymore!

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: Thank you for the information, I removed the function from the parser.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: It is not used by any other function as far as I can see now. I changed it to a class method for now, and will be cleaning up the parser structure in the end to address issues like this.

JFRudzinski:
@ladinesa Just a reminder in case you lost track of this one. Could you give this code a review whenever you have a chance (no rush). Thanks!

ladinesa:
sorry a bit confused, I implemented an initial parser for lammps and it was extended and merged. This pr aims to extend  and allign with the changes in the model system  defs. But seeing the changes, it seems that it is introducing a new parser. Can you enlighten me a bit?

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+
+from ..utils.mdanalysisparser import MDAnalysisParser
+from ..utils.mdparserutils import MDParser
+from nomad.units import ureg
+from nomad_simulations.schema_packages.general import Program, Simulation
+from nomad_simulations.schema_packages.atoms_state import ParticleState
+from nomad_simulations.schema_packages.model_method import ModelMethod
+from nomad_simulations.schema_packages.model_system import Cell, ModelSystem
+
+from simulationworkflowschema import (
```
ladinesa: I suggest you use the workflow defs in nomad-simulations

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+    GeometryOptimization,
+    GeometryOptimizationMethod,
+    GeometryOptimizationResults,
+)
+from simulationworkflowschema.molecular_dynamics import (
```
ladinesa: oh, this one ok, maybe use the old schema but please create an issue to change this to the new schema

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+    get_bond_list_from_model_contributions,
+)
+from structlog.stdlib import BoundLogger
+
+re_float = r'[-+]?\d+\.*\d*(?:[Ee][-+]\d+)?'
```
ladinesa: maybe put this in utils? or make use all caps `RE_FLOAT`

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+re_n = r'[\n\r]'
+
+
+def get_unit(units_type, property_type=None, dimension=3):
```
ladinesa: specify types for args and return value

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+    mole = 6.022140857e23
```
ladinesa: put in utils?

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+class TrajParser(TextParser):
+    def __init__(self):
+        self._masses = None
+        self._reference_masses = dict(
+            masses=np.array(asedata.atomic_masses), symbols=asedata.chemical_symbols
+        )
+        self._chemical_symbols = None
+        super().__init__(None)
+
+    def get_pbc_cell(self, val) -> tuple[list, np.ndarray]:
```
ladinesa: specify type for arg

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        else:  # TODO: orthogonal can be ff or ss (or mm?)
+            pbc = [v == 'pp' for v in val[:3]]
+            for i in range(3):
+                cell[i][i] = float(val[i * 2 + 4]) - float(val[i * 2 + 3])
+
+        return pbc, cell
+
+    def init_quantities(self):
+        def get_atoms_info(val):
+            val = val.split('\n')
```
ladinesa: use splitlines()

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+
+    @property
+    def with_trajectory(self) -> bool:
+        return self.get('atoms_info') is not None
+
+    @property
+    def n_frames(self) -> int:
+        return len(self.get('atoms_info', []))
+
+    @property
```
ladinesa: why make this as property when there is no update logic? maybe simply have self.masses or an update logic like `if self._masses is None: ....`

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+    def masses(self) -> np.ndarray:
+        return self._masses
+
+    # TODO: handle non-atomistic representations
+    @masses.setter
```
ladinesa: ah you have a setter, forget my other comment

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+    def masses(self, val):
```
ladinesa: type for arg

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        self._masses = val
```
ladinesa: shouldnt this be ```
if val is None:
    return```

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        if self._masses is None:
+            return
+
+        self._masses = val
```
ladinesa: put a sanity check on val before setting, in line 301 you are assuming it to be at least a 2d array

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+    if property_type:
+        return units.get(property_type, None)
+    else:
+        return units
+
+
+class TrajParser(TextParser):
+    def __init__(self):
+        self._masses = None
+        self._reference_masses = dict(
```
ladinesa: I suggest you make masses and chem symbols as global vars or include it in utils

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+            self._chemical_symbols = {}
+            for i in range(len(masses)):
+                symbol_idx = np.argmin(
+                    abs(self._reference_masses['masses'] - masses[i][1])
+                )
+                self._chemical_symbols[masses[i][0]] = self._reference_masses[
+                    'symbols'
+                ][symbol_idx]
+
+    def get_atom_labels(self, idx):
```
ladinesa: type for arg and return val

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        atoms_info = self.get('atoms_info')
+        if atoms_info is None:
+            return
+
+        atoms_id = atoms_info[idx].get('id')
```
ladinesa: more robust: `atoms_info.get(idx, {}).get('id')`

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        default = ['CGX' for _ in atoms_id] if atoms_id is not None else None
+        atoms_type = atoms_info[idx].get('type')
```
ladinesa: maybe define`id_info = atoms_info.get(idx)` at the top

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        if atoms_type is None:
+            return default
+        if self._chemical_symbols is None:
```
ladinesa: shouldn't this be at the start? oh you determine default, sorry

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+            return default
+
+        try:
+            atom_labels = [self._chemical_symbols[atype] for atype in atoms_type]
+        except Exception:
+            self.logger.error('Error resolving atom labels.')
+            return
```
ladinesa: maybe the return val is [] or simply set atom_labels = [] actually better you do not need the try as there seems to be exceptions to be had just do `self._chemical_symbols.get(atype)` then filter return [label for label in atom_labels if label is not None]

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+
+        return atom_labels
+
+    def get_positions(self, idx):
```
ladinesa: arg and return type

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        atoms_info = self.get('atoms_info')
+        if atoms_info is None:
+            return
+
+        atoms_info = atoms_info[idx]
+
+        cell = self.get('pbc_cell')
+        cell = None if cell is None else cell[idx][1]
```
ladinesa: shouldnt you have `if is cell is None:
return` here?

ladinesa: ah you have the l349 case, put this as the first case then all that requires cell in elif and return if cell is None

ladinesa: just in general I discourage using the return statement right away as it makes extending the code a little trickier, set instead the return value to something and have only one return statement at the end

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+                    [atoms_info['ix'], atoms_info['iy'], atoms_info['iz']]
+                ).T
+
+                positions += positions_img * np.linalg.norm(cell, axis=1)
+        else:
+            positions = None
+
+        return positions
+
+    def get_velocities(self, idx):
```
ladinesa: type for arg and return val

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        atoms_info = self.get('atoms_info')
```
ladinesa: this is becoming so repetitive I suggest having a common function for this.

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        if atoms_info is None:
+            return
+        atoms_info = atoms_info[idx]
+        if 'vx' not in atoms_info or 'vy' not in atoms_info or 'vz' not in atoms_info:
+            return
+
+        return np.array([atoms_info['vx'], atoms_info['vy'], atoms_info['vz']]).T
```
ladinesa: np.transpose([atoms_info.get(key) for key in (vs, vy, vz)])

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+
+    def get_forces(self, idx):
```
ladinesa: type for arg and return val

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        atoms_info = self.get('atoms_info')
+        if atoms_info is None:
+            return
+        atoms_info = atoms_info[idx]
+        if 'fx' not in atoms_info or 'fy' not in atoms_info or 'fz' not in atoms_info:
+            return
+        return np.array([atoms_info['fx'], atoms_info['fy'], atoms_info['fz']]).T
```
ladinesa: I suggest having a common function for get_forces and get_velocities e.g. get_atoms_array? that take in the key either f or v.

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+
+    def get_lattice_vectors(self, idx):
```
ladinesa: arg and return val type

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        pbc_cell = self.get('pbc_cell')
+        if pbc_cell is None:
+            return
+        return pbc_cell[idx][1]
```
ladinesa: is pbc_cell a dict? return self.get('pbc_cell', {}).get(idx, [None])[0]

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+
+    def get_pbc(self, idx):
```
ladinesa: have a common function for get_lattice_vectors and get_pbc taking in 0 or 1 as arg.

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        pbc_cell = self.get('pbc_cell')
+        if pbc_cell is None:
+            return
+        return pbc_cell[idx][0]
+
+    def get_n_atoms(self, idx):
```
ladinesa: type for arg and return val

ladinesa: have a common function for get_n_atoms and get_step taking in key and idx

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+
+    def get_step(self, idx):
+        step = self.get('time_step')
+        if step is None:
+            return
+        return step[idx]
+
+
+class XYZTrajParser(TrajParser):
+    def __init__(self):
```
ladinesa: you do not need this if there is nothing to init

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        super().__init__()
+
+    def init_quantities(self):
+        def get_atoms_info(val_in):
+            val = [v.split('#')[0].split() for v in val_in.strip().split('\n')]
```
ladinesa: use splitlines()

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+            symbols = []
+            for v in val:
+                if v[0].isalpha():
+                    if v[0] not in symbols:
+                        symbols.append(v[0])
+                    v[0] = symbols.index(v[0]) + 1
+            val = np.transpose(np.array([v for v in val if len(v) == 4], dtype=float))
+            # val[0] is the atomic number
+            val[0] = [list(set(val[0])).index(v) + 1 for v in val[0]]
+            return dict(type=val[0], x=val[1], y=val[2], z=val[3])
```
ladinesa: more pythonic return {key: val[n] for n, key in enumerate(['type', 'x', 'y', 'z'])}, but as it stands i'm fine with it.

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+                r'((?:\d+|[A-Z][a-z]?) [\s\S]+?)(?:\s\d+\n|\Z)',
+                str_operation=get_atoms_info,
+                comment='#',
+                repeats=True,
+            )
+        ]
+
+
+class TrajParsers:
+    def __init__(self, parsers):
```
ladinesa: type for arg

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        self._parsers = parsers
+        for parser in parsers:
+            parser.parse()
+
+    def __getitem__(self, index):
```
ladinesa: type for arg and return val

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        if self._parsers:
+            return self._parsers[index]
+
+    def eval(self, key, *args, **kwargs):
```
ladinesa: - type for arg and return val

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        ]
+        super().__init__(None)
+
+    def init_quantities(self):
+        self._quantities = [
+            Quantity(header, rf'{re_n} *(\d+) +{header}', repeats=True, dtype=np.int32)
+            for header in self._headers
+        ]
+
+        def get_section_value(val):
```
ladinesa: type for arg and return al

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+            'angle',
+            'dihedral',
+            'improper',
+            'kspace',
+        ]
+        self._units = None
+        super().__init__(None)
+
+    def init_quantities(self):
+        def str_op(val):
```
ladinesa: type for arg and return val

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+            val = val.split('#')[0]
+            val = val.replace('&\n', ' ').split()
```
ladinesa: you also have \r as escape char?

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+            val = val if len(val) > 1 else val[0]
+            return val
+
+        self._quantities = [
+            Quantity(
+                name,
+                r'\n\s*%s\s+(?!.*\$\{)([${}\w\. \/\#\-]+)(\&\n[\w\. \/\#\-]*)*' % name,
```
ladinesa: use f-string format

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+
+        self._quantities.append(
+            Quantity(
+                'minimization_stats',
+                r'\s*Minimization stats:\s*([\s\S]+?)\n\n',
+                flatten=False,
+            )
+        )
+
+        def str_to_thermo(val):
```
ladinesa: arg and return val type

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+            res = {}
+            if val.count('Step') > 1:
+                val = (
+                    val.replace('--', '').replace('=', '').replace('(sec)', '').split()
```
ladinesa: you can do a regex sub instead

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+                )
+                val = [v.strip() for v in val]
+
+                for i in range(len(val)):
+                    if val[i][0].isalpha():
+                        res.setdefault(val[i], [])
+                        res[val[i]].append(float(val[i + 1]))
+
+            else:
+                val = val.split('\n')
```
ladinesa: splitlines

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        )
+
+    @property
+    def units(self):
+        if self._units is None:
+            units_type = self.get('units', ['lj'])[0]
+            self._units = get_unit(units_type)
+        return self._units
+
+    def get_thermodynamic_data(self):
```
ladinesa: return val type

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+                data[key] = val * self.units.get('energy', 1)
+            elif low_key == 'press':
+                data[key] = val * self.units.get('pressure', 1)
+            elif low_key == 'temp':
+                data[key] = val * self.units.get('temperature', 1)
+            else:
+                data[key] = val
+        return data
+
+    def get_traj_files(self):
```
ladinesa: return val type

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        dump = self.get('dump', None)
+        if dump is None:
+            self.logger.warning('Trajectory not specified in directory, will scan.')
+            # TODO: extend matching of traj file
+            traj_files = os.listdir(self.maindir)
+            traj_files = [
+                f
+                for f in traj_files
+                if f.endswith('trj') or f.endswith('xyz')
```
ladinesa: use insead if f.split('.')[-1] in supported_traj files

ladinesa: we have utils.general.search_files, you should use this

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+            if type(dump[0]) in [str, int]:
+                dump = [dump]
+            traj_files = [d[4] for d in dump]
+        traj_files = [
+            i for n, i in enumerate(traj_files) if i not in traj_files[:n]
+        ]  # remove duplicates
+
+        return [os.path.join(self.maindir, f) for f in traj_files]
+
+    def get_data_files(self):
```
ladinesa: type for return val

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        def check_file_header(file_path, regex_pattern):
```
ladinesa: type for arg and return val

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        read_data = self.get('read_data')
+        # TODO: Check in with Leo on the reasoning behind this change
+        # # Chop out 'CPU' before, then just check none
+        # if read_data is not None:
+        #     try:
+        #         read_data.remove('CPU')
+        #     except Exception:
+        #         pass
+        if read_data is None or 'CPU' in read_data:
+            self.logger.warning('Data file not specified in directory, will scan.')
```
ladinesa: define this warning  in init, oh different from Traj file

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+            data_files = os.listdir(self.maindir)
```
ladinesa: use search_files util

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+                ]
+        else:
+            data_files = read_data
+
+        if not data_files:
+            self.logger.warning('No data_files found to match the log file.')
+
+        return [os.path.join(self.maindir, f) for f in data_files]
+
+    def get_pbc(self):
```
ladinesa: return val type

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        pbc = self.get('boundary', ['p', 'p', 'p'])
+        return [v == 'p' for v in pbc]
+
+    def get_sampling_method(self):
```
ladinesa: retun val type

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        fix_style = self.get('fix', [[''] * 3])[0][2]
+
+        sampling_method = (
+            'langevin_dynamics' if 'langevin' in fix_style else 'molecular_dynamics'
+        )
+        return sampling_method, fix_style
+
+    def get_thermostat_settings(self):
```
ladinesa: return val type

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        fix = self.get('fix', [None])[0]
```
ladinesa: instead of default [None] can be instead [[]] so you do not need l916-917

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+
+        else:
+            self.logger.warning('Fix style not supported', data=dict(style=fix_style))
+
+        return res
+
+    def get_interactions(self):
+        styles_coeffs = []
+        for interaction in self._interactions:
+            styles = self.get('%s_style' % interaction, None)
```
ladinesa: use f-format

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        return styles_coeffs
+
+
+class LammpsArchiveWriter(MDParser):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.log_parser = LogParser()
+        self.units = self.log_parser.units
+        self.aux_log_parser = LogParser()
+        self._md_parser = MDParser()
```
ladinesa: be consistent if all parsers ought to be private

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        self._traj_parser = TrajParser()
+        self._xyztraj_parser = XYZTrajParser()
+        self._mdanalysistraj_parser = MDAnalysisParser(
+            topology_format='DATA', format='LAMMPSDUMP'
+        )
+        self.data_parser = DataParser()
+
+    def apply_unit(self, value, unit):
```
ladinesa: arg and return val type

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+        if not hasattr(value, 'units'):
+            value = value * self.units.get(unit, 1)
+        return value
+
+    def parse_method(self, simulation):
```
ladinesa: why need to pass simulation?

ladinesa: it seems nothing is set in this function, can you describe what is really going on here.

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+    #         neighmodify = val[0]  # just use the first instace for now
+    #         neighmodify = np.array([str(i).lower() for i in neighmodify])
+    #         if 'every' in neighmodify:
+    #             index = np.where(neighmodify == 'every')[0]
+    #             sec_neighbor_searching.neighbor_update_frequency = int(
+    #                 neighmodify[index + 1]
+    #            )
+
+    def parse_system(self, simulation):
+        # sec_run = self.archive.run[-1]
```
ladinesa: remove

src/nomad_simulation_parsers/parsers/lammps/parser.py
ladinesa: i am confused, LammpsAchiveWriter is an MDParser but you have _md_parser

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+                },
+                'labels': self.traj_parsers.eval('get_atom_labels', traj_n),
+                'n_particles': self.traj_parsers.eval('get_n_atoms', traj_n),
+                'positions': self.apply_unit(
+                    self.traj_parsers.eval('get_positions', traj_n), 'distance'
+                ),
+                'velocities': velocities,
+                'bond_list': self._bond_list if self._bond_list else None,
+            }
+            self._md_parser.parse_trajectory_step(particles_dict, simulation)
```
ladinesa: just do self.parser_trajectory_step here

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+    #     if val is not None:
+    #         neighmodify = val[0]  # just use the first instace for now
+    #         neighmodify = np.array([str(i).lower() for i in neighmodify])
+    #         if 'every' in neighmodify:
+    #             index = np.where(neighmodify == 'every')[0]
+    #             sec_neighbor_searching.neighbor_update_frequency = int(
+    #                 neighmodify[index + 1]
+    #            )
+
+    def parse_system(self, simulation):
```
ladinesa: this is just too long, organize the contents and split them in chunks so it is easier to read

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+            ]
+            if 'CGX' in atoms_elements:
+                atoms_elements = (
+                    np.array(atom_labels)
+                    if atom_labels and 'CGX' not in atom_labels
+                    else atoms_types
+                )
+            atoms_resnames = np.array(atoms_info.get('resnames', []))
+            moltypes = np.unique(atoms_moltypes)
+            # for i_moltype, moltype in enumerate(moltypes):
```
ladinesa: what is going on here?

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+            #             # get the index of the first atom of each residue
+            #             ids_firstatom = np.cumsum(ids_count)[:-1]
+            #             # add the 0th index manually
+            #             ids_firstatom = np.insert(ids_firstatom, 0, 0)
+            #             names_firstatom = names[ids_firstatom]
+            #             sec_molecule.composition_formula = get_composition(
+            #                 names_firstatom
+            #             )
+
+    def write_to_archive(self) -> None:
```
ladinesa: this is just too long, organize it into smaller chunks

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+            self.logger.warning('Multiple data files are specified')
+        if data_files:
+            self.data_parser.mainfile = data_files[0]
+
+        # parse trajectorty file associated with calculation
+        traj_files = self.log_parser.get_traj_files()
+        if len(traj_files) > 1:
+            self.logger.warning('Multiple traj files are specified')
+
+        parsers = []
```
ladinesa, line 271: you can have this as a separate function get_traj_parsers

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+                self.log_parser.maindir, self.log_parser.get('log')[0]
+            )
+            # we assign units here which is read from log parser
+            self.aux_log_parser._units = self.log_parser.units
+
+        self.parse_method(self.archive.data)
+
+        self.parse_system(self.archive.data)
+
+        # # include input controls from log file
```
ladinesa, line 285: either put TODO or remove

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,1485 @@
...
+
+class LammpsParser(MatchingParser):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.archive_writer = LammpsArchiveWriter()
+
+    # ? Really needed for the LAMMPS parser?
+    # ? Would it make sense to handle a potential auxillary log file here,
+    # ? since there seems to be an issue with its handling in the old parser?
+    def is_mainfile(
```
ladinesa: remove, only need this for complex workflows

tests/parsers/test_lammps_parser.py
```diff
@@ -0,0 +1,384 @@
...
+#     )
+
+#     sec_sccs = archive.run[0].calculation
+#     assert len(sec_sccs) == 301
+#     assert sec_sccs[98].energy.total.value.magnitude == approx(1.45322428e-17)
+
+#     assert len(archive.run[0].system) == 4
+
+
+def test_traj_xyz(parser):
```
ladinesa: this should only use the XYZ parser not the main parser

tests/parsers/test_lammps_parser.py
```diff
@@ -0,0 +1,384 @@
...
+        'tests/data/lammps/methane_xyz/log.methane_nvt_traj_xyz_thermo_style_custom',
+        archive,
+        LOGGER,
+    )
+    sec_systems = archive.data.model_system
+    assert len(sec_systems) == 201
+    assert sec_systems[13].positions[7][0].magnitude == pytest.approx(-8.00436e-10)
+
+
+def test_traj_dcd(parser):
```
ladinesa: same here

tests/parsers/test_lammps_parser.py
```diff
@@ -0,0 +1,384 @@
...
+                for particle_state in sec_systems[107].particle_states
+            ]
+        )
+        == 320
+    )
+
+
+def test_unwrapped_pos(parser):
+    archive = EntryArchive()
+    parser.parse('tests/data/lammps/1_xyz_files/log.lammps', archive, LOGGER)
```
ladinesa: you only parse the file containing the pos and use only the relevant parser

tests/parsers/test_lammps_parser.py
```diff
@@ -0,0 +1,384 @@
...
+from nomad.utils import get_logger
+from nomad_simulation_parsers.parsers.lammps.parser import (
+    LammpsParser,
+    TrajParser,
+    TrajParsers,
+)
+
+LOGGER = get_logger(__name__)
+
+
```
ladinesa, line 40: organize your test in the same way you organize your classes in the parser e.g.
```
class TestXYYParser:
 ...

class LogParser:
...

class TestLammpsArchiveWriter:
...

```

and test individual functionalities.

ladinesa:
Please split and organize the file parsers into several files, see the other parsers in this repo. I did not look the tests in greater detail but please organize the tests in the same way you organize your classes. Also remove all comments that were taken from the old parser. just put a TODO e.g. in tests

tests/parsers/test_lammps_parser.py
Bernadette-Mohr: While I'm at it changing the tests anyway, would you recommend replacing the test data files with fixtures/mock data/temporary files?

tests/parsers/test_lammps_parser.py
Bernadette-Mohr: Done, as far as the structure is clear so far. I'll improve on it as I go.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: I've moved those to `utils.constants_definitions` for now. If acceptable, I will update their use in other parsers.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: Absolutely! It's changed now.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: I have moved the definition to `utils.constants_definitions` as well.

I am wondering why we aren't using `ureg.mole` here instead of carrying an additional definition of Avogadro's number through all parsers I am familiar with? Will test if using any of the existing `pint` definitions would work.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: I've added a check for `val` being a numpy array of the appropriate dimension. Please suggest a more elegant solution if mine isn't optimal!

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: `REFERENCE_MASSES` and `CHEMICAL_SYMBOLS` (both internal definitions adopted from ASE) are defined as constants in `utils.constants_definitions` now.

Also, I separated the derivation of system-applicable chemical symbols from the masses setter.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: I also changed the syntax for `atoms_type` accordingly.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: I've changed the flow a bit, to improve readability and minimize redundant checks for the availiability of `cell`. There's only one final `return` at the end now.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: I created helper functions for accessing `atoms_info` and coordinates, and restructured `get_positions`, `get_velocities`, and `get_forces`.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: `_get_frame_atoms_info` and `_extract_vector_components` are now used by `get_positions`, `get_velocities`, and `get_forces`.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: `pbc_cell` is a list of tuples, so no `get`. I have increased the defensiveness of the new `_get_cell_component` function.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: See comment above, there is `_get_cell_component` now.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: Both functions use `_get_trajectory_info` now.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: I find your suggestion more readable, thank you!

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: Absolutely. Using the already existing constant `RE_N` for handling different escape chars.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: I've made the change, still need to test it.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: Do you mean it is ok to define this warning here, since it is for Lammps data files specifically, or would you still prefer both data- and trajectory file warning messages definied in the constructor?

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: @JFRudzinski is currently working on migrating `molecular_dynamics`, I will update the dependencie accordingly.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: There's nothing set in this function yet, I prioritized porting `parse_system` to the new schema for testing. I'll get to `parse_method` as soon as I'm finished with the system.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: I am not sure why we cover the case of "no dump command in the log file" when identifying trajectory files? Afaik, without an explicit dump command, Lammps wouldn't ouput a trajectory in the first place.
But since there probably is a good reason for this logic being here, I extended the scanning capablility by `find_best_matching_file` to handle the fact that LAMPPS seems to enforce no file naming defaults or rules whatsoever.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: Agreed. I am currently reorganizing it on branch [`lammps-parser_system-hierarchy`](https://github.com/FAIRmat-NFDI/nomad-parser-plugins-simulation/pull/47) that holds the complete updated `parse_system`.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: Nothing further here on this PR, updating parsing the system hierarchy was decoupled and moved to [`lammps-parser_system-hierarchy`](https://github.com/FAIRmat-NFDI/nomad-parser-plugins-simulation/pull/47)

src/nomad_simulation_parsers/parsers/lammps/parser.py
JFRudzinski: ureg.mole is a unit, this number is a constant, IMO, badly named, it should be NA (i.e., Avogadro's number)

JFRudzinski (commented):
@Bernadette-Mohr I suggest making a small TODO statement if there are any small uncertainties left from Alvin's extensive review, and then resolve those comments. If there are any comments that you cannot make a single-line TODO to double check later, then tag Alvin again so it is clear what he should check.

I think these will get naturally addressed as you fill the later sections of the parser code and if not you can ensure that they get tackled in a later PR where there is not so much noise from other issues.

When you are done with that and fix the formatting I will check and approve.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: True, ureg.avogadro_constant or ureg.avogadro_number are probably closer to what we need, but they are already defined as well.

src/nomad_simulation_parsers/parsers/lammps/file_parsers.py
```diff
@@ -0,0 +1,771 @@
...
+
+            value = []
+            for i in range(len(val)):
+                v = val[i].split('#')[0].split()
+                if not v:
+                    continue
+
+                try:
+                    value.append(np.array(v, dtype=float))
+                # ? Should we append None here, instead of leaving value potentially an empty list?
```
ladinesa: you can put a return name, None in l277 instead of break

src/nomad_simulation_parsers/parsers/lammps/file_parsers.py
```diff
@@ -0,0 +1,771 @@
...
+                coeffs = list(coeffs)
+
+            styles_coeffs += coeffs
+
+        return styles_coeffs
+
+
+class LogParser(TextParser):
+    # TODO: add parsers for other LAMMPS-supported trajectory formats
+    SUPPORTED_TRAJ_EXTENSIONS = ['trj', 'xyz', 'dcd', 'lammpstrj']  # , 'h5', 'nc'
```
ladinesa: normally, only global vars are all caps, I suggest `_supported_traj_extensions` and `_data_file_patterns`

src/nomad_simulation_parsers/parsers/lammps/file_parsers.py
```diff
@@ -0,0 +1,771 @@
...
+    DATA_FILE_PATTERNS = ['data', 'dat']
+
+    def __init__(self) -> None:
+        self._commands = [
+            'angle_coeff',
```
ladinesa, line 313: these lists are becoming very long, do you think we will have to extend later. Can you please make a list of these extra long lists and maybe we can put them in a `config.json`

src/nomad_simulation_parsers/parsers/lammps/file_parsers.py
```diff
@@ -0,0 +1,771 @@
...
+                convert=False,
+            )
+        )
+
+    @property
+    def units(self) -> dict[str, pint.Quantity]:
+        if self._units is None:
+            units_type = self.get('units', ['lj'])[0]
+            self._units = get_unit(units_type)
+        return self._units
```
ladinesa, line 533: please add this so we reset units and potential other private vars when we set log file:
```python
def reset(self):
    super().reset()
    self._units = None
```

src/nomad_simulation_parsers/parsers/lammps/file_parsers.py
```diff
@@ -0,0 +1,771 @@
...
+        return matching_files
+
+    def get_traj_files(self) -> list[str]:
+        dump = self.get('dump', None)
+        if dump is None:
+            self.logger.warning('Trajectory not specified in directory, will scan.')
+            traj_files = []
+            for ext in self.SUPPORTED_TRAJ_EXTENSIONS:
+                found_files = search_files(
+                    pattern=f'*{ext}*',
```
ladinesa: should be f`*.{ext}`?

src/nomad_simulation_parsers/parsers/lammps/file_parsers.py
```diff
@@ -0,0 +1,771 @@
...
+            ]
+
+            def get_tokens(filename: str) -> set[str]:
+                """Extract tokens from filename by splitting on delimiters."""
+                basename = os.path.basename(filename).rsplit('.', 1)[0].lower()
+                tokens = re.split(r'[_\-\.\d]+', basename)
+                MIN_CHARS = 2
+                return {t for t in tokens if len(t) > MIN_CHARS}
+
+            def get_best_from_scores(
```
ladinesa, line 585: there is this [module](https://docs.python.org/3/library/difflib.html#difflib.get_close_matches) maybe you can use it to sort the files or [bisect](https://docs.python.org/3/library/bisect.html). But looks good already.

src/nomad_simulation_parsers/parsers/lammps/file_parsers.py
```diff
@@ -0,0 +1,771 @@
...
+                        traj_file,
+                    )
+                    for traj_basename, traj_file in traj_data
+                ]
+                matching_files = get_best_from_scores(sequence_scores, 0.3)
+
+            # ? Looking for the biggest file size/keywords (result/final/dry ...) could make more sense?
+            # Fallback: Use first file
+            if not matching_files:
+                self.logger.warning(
```
ladinesa, line 623: do not use formatted string

src/nomad_simulation_parsers/parsers/lammps/file_parsers.py
```diff
@@ -0,0 +1,771 @@
...
+            return {}
+
+        temp_unit = self.units.get('temperature', 1)
+        press_unit = self.units.get('pressure', 1)
+        time_unit = self.units.get('time', 1)
+
+        res = dict()
+        if fix_style.lower() == 'nvt':
+            try:
+                res['target_T'] = float(fix[5]) * temp_unit
```
ladinesa: if possible, convert fix to list of floats so we do not see these float() conversions

src/nomad_simulation_parsers/parsers/lammps/file_parsers.py
```diff
@@ -0,0 +1,771 @@
...
+            if isinstance(styles[0], str):
+                styles = [styles]
+
+            for i in range(len(styles)):
+                if interaction == 'kspace':
+                    coeff = [[float(c) for c in styles[i][1:]]]
+                    style = styles[i][0]
+
+                else:
+                    coeff = self.get('%s_coeff' % interaction)
```
ladinesa: use f'' format

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,542 @@
...
+    def parse_method(self, simulation: Simulation) -> None:
+        # TODO: replace with counterparts from nomad_simulations!
+
+        if self.traj_parsers[0].mainfile is None or self._data_parser.mainfile is None:
+            return
+
+        if self.traj_parsers.eval('n_frames') is None:
+            return
+
+        sec_method = ModelMethod()
```
ladinesa: please remove l49-154 so we have a clean code, we add them later in a separate mr

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,542 @@
...
+            ]
+            if 'CGX' in atoms_elements:
+                atoms_elements = (
+                    np.array(atom_labels)
+                    if atom_labels and 'CGX' not in atom_labels
+                    else atoms_types
+                )
+            atoms_resnames = np.array(atoms_info.get('resnames', []))
+            moltypes = np.unique(atoms_moltypes)
+            # for i_moltype, moltype in enumerate(moltypes):
```
ladinesa: please remove all commented lines add them in a separate mr

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,542 @@
...
+            #             ids_firstatom = np.insert(ids_firstatom, 0, 0)
+            #             names_firstatom = names[ids_firstatom]
+            #             sec_molecule.composition_formula = get_composition(
+            #                 names_firstatom
+            #             )
+
+    def _configure_parsers(self) -> None:
+        """Configure all parsers with loggers and basic settings."""
+
+        def _setup_auxiliary_log_parser(aux_log_name) -> None:
```
ladinesa: you do not need to define this function, simply put the code to the if statement below.

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,542 @@
...
+        # Configure trajectory parsers
+        self._traj_parser.logger = self.logger
+        self._traj_parser._chemical_symbols = None
+        self._xyztraj_parser.logger = self.logger
+        self._mdanalysistraj_parser.logger = self.logger
+
+        # Configure data parser
+        self._data_parser.logger = self.logger
+
+    def _parse_data_files(self) -> None:
```
ladinesa: is this really _parse_ or _set_?

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,542 @@
...
+        """Parse and configure data file(s) associated with calculation."""
+        data_files = self._log_parser.get_data_files()
+
+        if len(data_files) > 1:
+            self.logger.warning('Multiple data files are specified')
+
+        if data_files:
+            self._data_parser.mainfile = data_files[0]
+
+    def _create_formatted_parser(
```
ladinesa: this is not used outside the scope of _create_trajectory_parser put it there if linting permits

ladinesa: or maybe not posible as you have long _create_custom_parser.

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,542 @@
...
+        self, traj_file: str, index: int, data_file: str
+    ) -> TrajParser | XYZTrajParser | MDAnalysisParser:
+        """
+        Create appropriate trajectory parser based on file type.
+
+        Parser initialization for each traj file cannot be avoided as there are
+        cases where traj files can share the same parser.
+        """
+
+        def _get_trajectory_file_type(traj_file: str, index: int) -> str:
```
ladinesa: you do not need to define this, you run into a confusing code with all the sub-sub-funcs

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,542 @@
...
+
+    def _create_fallback_parser(self, traj_file: str) -> TrajParser:
+        """Create fallback TrajParser when file type is not recognized."""
+        self.logger.warning(f'File type of {traj_file} not recognized.')
+        traj_parser = TrajParser()
+        traj_parser.mainfile = traj_file
+        # TODO: provide support for other file types
+        return traj_parser
+
+    def _create_trajectory_parser(
```
ladinesa, line 178: in general, if you can squeeze these in their relevant function defs please do so.

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,542 @@
...
+        ).get('names', []):
+            # MDAnalysis is necessary to calculate rdf and atomsgroup
+            if index == 0:
+                self._mdanalysistraj_parser = traj_parser
+            traj_parser = TrajParser()
+            traj_parser.mainfile = traj_file
+
+        return traj_parser
+
+    def _create_fallback_parser(self, traj_file: str) -> TrajParser:
```
ladinesa: you do not need to define this just insert the code if possible.

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,339 @@
...
+)
+from nomad_simulation_parsers.parsers.utils.mdanalysisparser import MDAnalysisParser
+
+
+class TrajParser(TextParser):
+    def __init__(self) -> None:
+        self._masses = None
+        self._chemical_symbols = None
+        super().__init__(None)
+
```
ladinesa, line 25: insert 
```
def reset(self):
    super().reset()
    self._masses = None
    self._chemical_symbols = None
```

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,339 @@
...
+    # TODO: handle non-atomistic representations
+    @masses.setter
+    def masses(self, val: Any) -> None:
+        if not val:
+            return
+        if not isinstance(val, np.ndarray):
+            try:
+                val = np.asarray(val)
+            except (ValueError, TypeError) as e:
+                raise ValueError(f'Cannot convert masses to array: {e}')
```
ladinesa: do not raise error log and return None instead

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,339 @@
...
+        min_dim = 2
+        if val.ndim < min_dim:
+            raise ValueError(f'masses must be at least 2D, got {val.ndim}D array')
```
ladinesa: same

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,339 @@
...
+        self._masses = val
+
+    @property
+    def chemical_symbols(self) -> dict | None:
+        """Chemical symbols derived from particle masses."""
+        if self._chemical_symbols is None and self._masses is not None:
+            self._derive_chemical_symbols()
+        return self._chemical_symbols
+
+    def _derive_chemical_symbols(self) -> None:
```
ladinesa: you do not need to define this. simply insert the code in l114

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,339 @@
...
+        """Derive chemical symbols from mass data."""
+        masses = self._masses[0][1]
+        self._chemical_symbols = {}
+        for i in range(len(masses)):
+            symbol_idx = np.argmin(abs(REFERENCE_MASSES - masses[i][1]))
+            self._chemical_symbols[masses[i][0]] = CHEMICAL_SYMBOLS[symbol_idx]
+
+    def get_atom_labels(self, idx: int) -> list[str] | None:
```
ladinesa: i suggest you return empty list instead of None and look at the usage. inspect also other parsers for consistency

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,339 @@
...
+        if atoms_type is None:
+            return default
+        if self._chemical_symbols is None:
+            return default
+
+        atom_labels = [self._chemical_symbols[atype] for atype in atoms_type]
+
+        return [label for label in atom_labels if label is not None]
+
+    def _get_frame_atoms_info(self, idx: int) -> dict | None:
```
ladinesa: maybe return an empty dict instead of None so you do not have the ifs

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,339 @@
...
+        def has_coords(*keys):
+            return all(k in frame_atoms_info for k in keys)
+
+        cell = self.get('pbc_cell')
+        cell = None if cell is None else cell[idx][1]
+
+        # Cell required
+        if cell is not None:
+            if has_coords('xs', 'ys', 'zs'):
+                positions = self._extract_vector_components(
```
ladinesa: check is positions is not None here

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,339 @@
...
+                    frame_atoms_info, 'xs', 'ys', 'zs'
+                )
+                positions = positions * np.linalg.norm(cell, axis=1) + np.amin(
+                    cell, axis=1
+                )
+            elif has_coords('xsu', 'ysu', 'zsu'):
+                positions = self._extract_vector_components(
```
ladinesa: also here

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,339 @@
...
+                    frame_atoms_info, 'xsu', 'ysu', 'zsu'
+                )
+                positions = positions * np.linalg.norm(cell, axis=1) + np.amin(
+                    cell, axis=1
+                )
+
+        # Unwrapped
+        if positions is None and has_coords('xu', 'yu', 'zu'):
+            positions = self._extract_vector_components(
```
ladinesa, line 193: and here

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,339 @@
...
+                frame_atoms_info, 'xu', 'yu', 'zu'
+            )
+
+        # Absolute positions with optional image correction
+        if positions is None and has_coords('x', 'y', 'z'):
+            positions = self._extract_vector_components(frame_atoms_info, 'x', 'y', 'z')
```
ladinesa, line 199: and here

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,339 @@
...
+from nomad.parsing.file_parser import Quantity, TextParser
+
+from nomad_simulation_parsers.parsers.utils.constants_definitions import (
+    CHEMICAL_SYMBOLS,
+    REFERENCE_MASSES,
+)
+from nomad_simulation_parsers.parsers.utils.mdanalysisparser import MDAnalysisParser
+
+
+class TrajParser(TextParser):
```
ladinesa, line 13: please add short description

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,339 @@
...
+                'atoms_info',
+                r'((?:\d+|[A-Z][a-z]?) [\s\S]+?)(?:\s\d+\n|\Z)',
+                str_operation=get_atoms_info,
+                comment='#',
+                repeats=True,
+            )
+        ]
+
+
+class TrajParsers:
```
ladinesa, line 292: add a short description

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,339 @@
...
+    def __init__(
+        self, parsers: list[TrajParser | XYZTrajParser | MDAnalysisParser]
+    ) -> None:
+        self._parsers = parsers
+        for parser in parsers:
+            parser.parse()
+
+    def __getitem__(self, index: int) -> TrajParser | XYZTrajParser | MDAnalysisParser:
+        if self._parsers is None or not self._parsers:
+            raise IndexError('No parsers available')
```
ladinesa: do not raise error log and return None

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,339 @@
...
+        return self._parsers[index]
```
ladinesa: you should put a try here instead of the if in l293

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,339 @@
...
+                found_attribute = True
+
+                if callable(parser_attr):
+                    # Always call if it's callable (even with no args)
+                    val = parser_attr(*args, **kwargs)
+                else:
+                    # It's a property/attribute - only return if no args expected
+                    if args or kwargs:
+                        self.logger.warning(
+                            f"Arguments provided for non-callable attribute '{key}'"
```
ladinesa: do not use formatted string

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,339 @@
...
+                        )
+                        continue
+                    val = parser_attr
+
+            except Exception as e:
+                self.logger.debug(
+                    f"Error evaluating '{key}' on {parser.__class__.__name__}: {e}"
```
ladinesa: also here

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,339 @@
...
+                )
+                continue
+
+        if not found_attribute:
+            raise AttributeError(f"Attribute '{key}' not found in parsers")
```
ladinesa: and here

src/nomad_simulation_parsers/parsers/utils/constants.py
ladinesa, line 1: maybe simply constants.py?

tests/parsers/test_lammps_parser.py
```diff
@@ -0,0 +1,814 @@
...
+    dcd_parser.logger = LOGGER
+    dcd_parser.parse()
+    # TODO: add assertion for calculation
+    positions = dcd_parser.get_positions(56)
+    assert np.shape(positions) == (320, 3)
+    labels = dcd_parser.get_atom_labels(107)
+    assert len(labels) == 320
+
+
+# TODO: re-include output testing alongside migration of parser functionalities
```
ladinesa: remove commented lines we add them later.

tests/parsers/test_lammps_parser.py
```diff
@@ -0,0 +1,814 @@
...
+                    'fy',
+                    'fz',
+                    'vx',
+                    'vy',
+                    'vz',
+                ]
+            ]
+        return default
+
+    parser.get = mock_get_with_dump
```
ladinesa: you can instead define 
```python 
class TestLogParser(LogParser):
    def parse(self, key):
        if key == 'dump':
            self._results['dump'] = ....
        else:
            super().parse(key)
```

tests/parsers/test_lammps_parser.py
```diff
@@ -0,0 +1,814 @@
...
+    assert len(data_files) == 1
+    assert os.path.basename(data_files[0]) == 'data.64xmethane_from_restart'
+    assert os.path.isabs(data_files[0])
+
+    # Test 2: Scan directory by pattern (no read_data command)
+    with tempfile.TemporaryDirectory() as tmpdir:
+        # Create test data files
+        data_file = os.path.join(tmpdir, 'data.polymer')
+        with open(data_file, 'w') as f:
+            f.write('LAMMPS data file. CGCMM style. atom_style full generated by VMD\n')
```
ladinesa: I do not prefer this file ops. Maybe try:
```
from io import StringIO
f = StringIO()
f.name = 'test'
f.write(....)
f.seek(0)
parser.mainfile = f
```

Bernadette-Mohr: We do a lot of directory scanning, so I don't think we can completely drop creating tempfiles.
But I reduced the number of file operations and limited the tests for `get_data_files` to only focus on `check_file_header` (not also used by `get_traj_files`) using your StringIO suggestion.
Let me know if I am now missing anything.

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,333 @@
...
+        self._log_parser = LogParser()
+        self._aux_log_parser = LogParser()
+        self._traj_parser = TrajParser()
+        self._xyztraj_parser = XYZTrajParser()
+        self._mdanalysistraj_parser = MDAnalysisParser(
+            topology_format='DATA', format='LAMMPSDUMP'
+        )
+        self._data_parser = DataParser()
+
+        self.units = self._log_parser.units
```
ladinesa: not necessary to define

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -0,0 +1,333 @@
...
+        # Set up and parse trajectory files
+        parsers = self._parse_trajectory_files()
+        if not self.traj_parsers or self.traj_parsers[0] is None:
+            return
+
+        # Parse system, method, parameters, thermodynamic data, etc.
+        self._parse_content_sections()
+
+        # Close all parser instances
+        self._mdanalysistraj_parser.close()
```
ladinesa, line 314: close also log and aux_log parsers

src/nomad_simulation_parsers/parsers/lammps/file_parsers.py
```diff
@@ -0,0 +1,799 @@
...
+            elif low_key == 'press':
+                data[key] = val * self.units.get('pressure', 1)
+            elif low_key == 'temp':
+                data[key] = val * self.units.get('temperature', 1)
+            else:
+                data[key] = val
+
+        return data
+
+    def find_best_matching_file(
```
ladinesa, line 560: this looks like a good candidate for a common utils function but could be moved in separate pr. maybe put a TODO

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,362 @@
...
+
+                found_attribute = True
+
+                if callable(parser_attr):
+                    # Always call if it's callable (even with no args)
+                    val = parser_attr(*args, **kwargs)
+                else:
+                    # It's a property/attribute - only return if no args expected
+                    if args or kwargs:
+                        self.logger.warning(
```
ladinesa, line 340: remove formatting

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,362 @@
...
+                            "Arguments provided for non-callable attribute '%s'", key
+                        )
+                        continue
+                    val = parser_attr
+
+            except Exception as e:
+                self.logger.debug(
```
ladinesa, line 347: also here

src/nomad_simulation_parsers/parsers/lammps/trajectory_parsers.py
```diff
@@ -0,0 +1,362 @@
...
+                    "Error evaluating '%s' on %s",
+                    key,
+                    parser.__class__.__name__,
+                    exc_info=e,
+                )
+                continue
+
+        if not found_attribute:
+            raise AttributeError("Attribute '%s' not found in parsers", key)
```
ladinesa: do not raise error

src/nomad_simulation_parsers/parsers/utils/constants.py
```diff
@@ -0,0 +1,9 @@
+import numpy as np
+from ase import data as asedata
+
+RE_FLOAT = r'[-+]?\d+\.*\d*(?:[Ee][-+]\d+)?'
+RE_N = r'[\n\r]'
+# TODO: Why not use ureg.avogadro_constant or ureg.avogadro_number from pint?
```
ladinesa: Not sure if your units would be an issue when setting your quantity. you can do maybe 1* ureg.avogadros_number.to_base_units(), can you maybe test this?

tests/parsers/test_lammps_parser.py
```diff
@@ -0,0 +1,586 @@
...
+    parser.logger = LOGGER
+    parser.mainfile = 'tests/data/lammps/1_xyz_files/log.test'
+    traj_files = parser.get_traj_files()
+
+    assert len(traj_files) == 1
+    assert os.path.basename(traj_files[0]) == 'pos_vel.xyz'
+    assert os.path.isabs(traj_files[0])
+
+    # Test 2: Scan directory (no dump command)
+    with tempfile.TemporaryDirectory() as tmpdir:
```
ladinesa: maybe make this as a fixture see for instance `nomad-FAIR/tests/conftest.py/tmp`

ladinesa:
This looks very good now! just have a couple minor issues you can merge it when resolved.

src/nomad_simulation_parsers/parsers/utils/constants.py
ladinesa: A quick check, it does work, no issue with setting quantity. you do not need to define this constant here then.
```
from nomad.metainfo import Quantity, MSection
from nomad.units import ureg
import numpy as np
class A(MSection):
    q = Quantity(type=np.float64, unit='joule')

a = A()
a.q = 1*ureg.kJ/ureg.avogadro_number
```

Bernadette-Mohr: Tested with the Lammps parser  `file_parsers.get_unit` function, `(1 * ureg.avogadro_number).to_base_units()` seems to work. We can discuss with @JFRudzinski if we want to replace the hardcoded constant in the parsers at some point.

I also realized that that value we use is outdated, so outsourcing the maintenance of constant definitions might really be something to think about.

JFRudzinski (approved):

PR #41, opened 2025-08-21
Updated parser_trajectory_step to new schema

Bernadette-Mohr:
I still do not have a good solution for the check for an existing archive, but otherwise, `parser_trajectory_step` now follows `ModelSystem`. `AtomState` is replaced by `ParticleState`, and `AtomicCell` is replaced by `Cell`.


JFRudzinski (approved):
you can merge this :+1:

PR #44, opened 2025-08-29
Fix import bug: path and classname

ndaelman-hu:
Patch in tandem with https://github.com/nomad-coe/nomad-simulations/pull/254

### DIFF
--- a/src/nomad_simulation_parsers/parsers/wannier90/parser.py
+++ b/src/nomad_simulation_parsers/parsers/wannier90/parser.py
@@ -15,7 +15,7 @@ from nomad.parsing.parser import MatchingParser
 from nomad.utils import get_logger
 from nomad_simulations.schema_packages.general import Simulation
 from nomad_simulations.schema_packages.workflow import SinglePoint
-from nomad_simulations.schema_packages.workflow.beyond_dft import DFTTBWorkflow
+from nomad_simulations.schema_packages.workflow.dmft import DFTTBDMFTWorkflow
 from structlog.stdlib import BoundLogger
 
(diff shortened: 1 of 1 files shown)


ndaelman-hu:
@Bernadette-Mohr @JFRudzinski `test_h5md_parser.py` keeps failing at line 87 `assert sec_atoms_group[0].is_molecule() is True`. Based on the `pytest` logs, my impression is that the `root` positions are not properly populated.

Bernadette-Mohr:
> @Bernadette-Mohr @JFRudzinski `test_h5md_parser.py` keeps failing at line 87 `assert sec_atoms_group[0].is_molecule() is True`. Based on the `pytest` logs, my impression is that the `root` positions are not properly populated.

From a cursory look, I think this might be a mistake in the test. If I undestand it correctly, a group of molecules (`branch_label == 'molecule_group'`) should not be identified as a molecule. I'm not sure though, will wait for @JFRudzinski .

ndaelman-hu:
> > @Bernadette-Mohr @JFRudzinski `test_h5md_parser.py` keeps failing at line 87 `assert sec_atoms_group[0].is_molecule() is True`. Based on the `pytest` logs, my impression is that the `root` positions are not properly populated.
> 
> From a cursory look, I think this might be a mistake in the test. If I undestand it correctly, a group of molecules (`branch_label == 'molecule_group'`) should not be identified as a molecule. I'm not sure though, will wait for @JFRudzinski .

Is very possible: the HDF5 file distinguishes between the branch and the molecule that it refers to. Idk whether this mapping is respected within NOMAD.

PR #46, opened 2025-09-03
Lammps parser system hierarchy

Bernadette-Mohr:
- [x] Polulate NOMAD archive with system hierarchy, according to new schema
- [x] Add tests for system parsing

As far as the available test data, the system hierarchy seems to be correctly parsed. I'd like to test a more complex system to be sure though.


pyproject.toml
```diff
@@ -34,6 +34,7 @@ license = { file = "LICENSE" }
 dependencies = [
     "nomad-lab>=1.3.16.dev100",
     "nomad-simulations@git+https://github.com/nomad-coe/nomad-simulations.git@develop",
+    "nomad-schema-plugin-simulation-workflow>=1.0.10",
```
JFRudzinski, line 37: I will check below if you are using this, but all workflow schema should now come directly from nomad-simulations

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -35,108 +37,418 @@ def apply_unit(self, value: Any, unit: str) -> float:
         return value
 
     def parse_method(self, simulation: Simulation) -> None:
-        # TODO: replace with counterparts from nomad_simulations!
+        """
+        Parse method information from data file and log file.
+
+        TODO: Migrate from legacy parser
```
JFRudzinski, line 43: move the TODO out of the docstring so that it is visible

tests/parsers/test_lammps_parser.py
```diff
@@ -584,8 +578,96 @@ def test_traj_xyz():
         os.unlink(temp_file)
 
 
+def test_systems(parser) -> None:
```
JFRudzinski, line 581: Can you open an issue that we need to reassess the efficiency of this test module in particular, but also create an overall strategy for parser testing, possibly with various levels of complexity...nothing to address in this PR, the tests are great as is for now

pyproject.toml
JFRudzinski: I removed this dependency and ran the tests which passed. Please remove.

JFRudzinski (approved):
Just some tiny comments

PR #47, opened 2025-09-04
Lammps parser system hierarchy

JFRudzinski:
@Bernadette-Mohr I am making a new PR here just for review of your system-hierarchy changes, so I can only see the latest changes and not the ones from the lammps-parser PR

In the end we can simply close this PR and merge directly into develop after the other PR is closed


pyproject.toml
```diff
@@ -33,7 +33,7 @@ maintainers = [
 license = { file = "LICENSE" }
 dependencies = [
     "nomad-lab>=1.3.16.dev100",
-    "nomad-simulations@git+https://github.com/nomad-coe/nomad-simulations.git@develop",
+    "nomad-simulations",  # @git+https://github.com/nomad-coe/nomad-simulations.git@develop",
```
JFRudzinski: remove changes

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -11,6 +11,7 @@
 
 from ..utils.mdanalysisparser import MDAnalysisParser
 from ..utils.mdparserutils import MDParser
+from nomad.client import normalize_all
```
JFRudzinski: remove

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -1435,6 +1410,8 @@ def write_to_archive(self) -> None:
 
         self.parse_system(self.archive.data)
 
+        # normalize_all(self.archive)
```
JFRudzinski: You shouldn't need this for testing, if you are running `nomad parse` it will run the normalization for you. If you are using the pytest module, then you should import and normalize there completely after parsin

Bernadette-Mohr:
> @Bernadette-Mohr I am making a new PR here just for review of your system-hierarchy changes, so I can only see the latest changes and not the ones from the lammps-parser PR
> 
> 
> 
> In the end we can simply close this PR and merge directly into develop after the other PR is closed

Thank you, that was exactly what I intended! I must have made a mistake when creating this PR.

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -1173,140 +1177,111 @@ def parse_system(self, simulation):
...
-            #             ids_firstatom = np.cumsum(ids_count)[:-1]
-            #             # add the 0th index manually
-            #             ids_firstatom = np.insert(ids_firstatom, 0, 0)
-            #             names_firstatom = names[ids_firstatom]
-            #             sec_molecule.composition_formula = get_composition(
-            #                 names_firstatom
-            #             )
+            particles_resnames = np.array(particles_info.get('resnames', []))
+            moltypes = np.unique(particles_moltypes)
+            simulation.model_system[0].dimensionality = 3
```
JFRudzinski: This should be set in the general parse_system section, when you get/set the cell info. Not sure what it was like before, but we shouldn't hard code it like this. We should either parse directly or infer from the parsed data (e.g., lattice vectors). I am sure that Lammps supports non-3D systems

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -1173,140 +1177,111 @@ def parse_system(self, simulation):
...
+            for i_moltype, moltype in enumerate(moltypes):
+                # Only add atomsgroup for initial system for now
```
JFRudzinski: This is a comment about only storing the hierarchy for the first frame in the trajectory, we should keep the approach and just rename to current schema

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -1173,140 +1177,111 @@ def parse_system(self, simulation):
...
+                # ! AtomsGroup deprecated, sub_system = ModelSystem() now!
```
JFRudzinski: Let's not refer to the old schema, it's a complete refactoring, so we have a clean break

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -1173,140 +1177,111 @@ def parse_system(self, simulation):
...
+                mol_nums = particles_molnums[sec_molecule_group.particle_indices]
+                moltype_count = np.unique(mol_nums).shape[0]
+                sec_molecule_group.composition_formula = f'{moltype}({moltype_count})'
+
+                molecules = particles_molnums
+                for i_molecule, molecule in enumerate(
+                    np.unique(molecules[sec_molecule_group.particle_indices])
+                ):
+                    sec_molecule = ModelSystem()
+                    # ? Does that happen automatically, or do I have to set 'is_representative'?
```
JFRudzinski: You have to set is_representative this if you want it to be the first frame, alternatively we could make the representative system the last frame (this will be done automatically if no representative system is set). However, I think in either case we should set it explicitly because we want it to be linked to whatever system we are attaching this hierarchy

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -1173,140 +1177,111 @@ def parse_system(self, simulation):
...
+                    # if i_molecule == 0:
+                    #     sec_molecule.is_representative = True
+                    sec_molecule.particle_indices = np.where(molecules == molecule)[0]
+                    # sec_molecule.n_particles = len(sec_molecule.particle_indices)
+                    sec_molecule.name = molecule
+                    sec_molecule.branch_label = 'molecule'
+                    # ? Set by normalizer?
```
JFRudzinski: there is no longer a quantity `is_molecule` but rather a class method that can be called. So, you don't need to set this

src/nomad_simulation_parsers/parsers/utils/mdparserutils.py
```diff
@@ -168,6 +168,7 @@ def parse_trajectory_step(
             model_system.particle_states.append(particle_state)
         self.parse_section(cell_dict, cell)
         model_system.cell.append(cell)
+        # ? Should the setting of model_system.dimensionality be done here?
```
JFRudzinski: yes, exactly, I think here, and derived from the parsed data somehow

JFRudzinski: yes, exactly, I think here, and derived from the parsed data somehow

tests/parsers/test_wannier90_parser.py
```diff
@@ -1,12 +1,12 @@
-from nomad.datamodel import EntryArchive
-from nomad.utils import get_logger
+# from nomad.datamodel import EntryArchive
```
JFRudzinski: Don't forget to revert this before merging, if it is broken, then we need to address it explicitly elsewhere

JFRudzinski (commented):
Looks good @Bernadette-Mohr ! I hope my comments help. After you make those changes, just clean up the comments and such

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: This was just a leftover from me printing out parsed test data to grab target values for the tests. It's removed.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: That's how I interpreted it, I just retained the comment. Extending this might be something we could think about for supporting reaction simulations or (semi) grand canonical ensemble simulations.

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: Of course, it's already removed. This was just a leftover of me mapping differences between the old and new schemas.

tests/parsers/test_wannier90_parser.py
Bernadette-Mohr: Reverted, that issue is already resolved.

src/nomad_simulation_parsers/parsers/utils/mdparserutils.py
Bernadette-Mohr: So far, the dimensionality isn't parsed anywhere.

Lammps supports 2D or 3D systems, I'm currently investigating if there is a flag in the logfile we could grab.

src/nomad_simulation_parsers/parsers/utils/mdparserutils.py
```diff
@@ -150,9 +150,8 @@ def parse_trajectory_step(
         """
         Create a system section and write the provided data.
         """
-        # ? How to handle a missing archive now?
-        # if self.archive is None:
-        #     return
+        if simulation is None:
+            return
```
Bernadette-Mohr, line 157: `simulation = self.archive.data`, so should work well enough here.

Bernadette-Mohr (commented):
I've tried to balance readability and minimal redundancy when restructuring `parse_system`. I could further reduce the number of helper functions with passing more keywords from e.g., a hierarchy key dictionary, but this would likely make the code harder to understand.

src/nomad_simulation_parsers/parsers/lammps/parser.py
```diff
@@ -1170,188 +1181,241 @@ def parse_system(self, simulation):
...
+        if particles_info is None:
+            return
+
+        particle_arrays = self._extract_particle_arrays(particles_info, simulation)
+
+        # Build molecular hierarchy
+        moltypes = np.unique(particle_arrays['moltypes'])
+        for moltype in moltypes:
+            molecule_group = self._create_molecule_group(moltype, particle_arrays)
+            simulation.model_system[0].sub_systems.append(molecule_group)
```
JFRudzinski, line 358: Since we are attaching the hierarchy only to system index 0, we should mark this as `is_representative` ... or maybe it is better to use the final frame? Idk

JFRudzinski (commented):
Looking very nice!

I think you should now update your branch and fix the ruff formatting...or what is your plan with the changes you made in the other PR, you merge that one first I guess?

src/nomad_simulation_parsers/parsers/lammps/parser.py
Bernadette-Mohr: I feel index 0 would be more consistent with what people intuitively expect.
Do you mean 'molecule_group' should be  `is_representative` here?

JFRudzinski: I meant that the root model_system should be representative. I thought maybe index -1 is better since some people think the last configuration is somehow indicative of something (i.e., post-equilibration)

JFRudzinski:
already in #46

PR #48, opened 2025-09-05
Qe gipaw parser

ladinesa:
Implementation of parsing of gipaw and efg outputs. Uses defs from nomad-nmr schema.


cecizag:
As requested by Alvin, here are the errors I encountered when testing this branch in my `nomad-distro-dev`.

* File included in the upload: only `quartz-scf.out`

  * the `process status` is `SUCCESS` and there are no errors in the `logs` section.

* Files included in the upload: `quartz-scf.out` + `quartz-nmr.out`

  * 3 entries are created, all with `process status` equal to `SUCCESS`
  * the entry of `type` `SimulationWorkflow` is empty, and this warning appears in the `logs`: 
  
image

  * the entry of `type` `Quantum Espresso DFT SinglePoint` associated with `quartz-scf.out` shows these errors in the `logs` section: 
  
image

  * the entry associated with `quartz-nmr.out` has no `type` assigned and shows these errors in the `logs` section:
  
image


* Files included in the upload: `quartz-scf.out` + `quartz-efg.out`

  * the situation is similar to the previous point.

* Files included in the upload: `quartz-scf.out` + `quartz-nmr.out` + `quartz-efg.out`

  * 5 entries are created, all with `process status` equal to `SUCCESS`
  * the entry of `type` `SimulationWorkflow` is the same as in the previous cases
  * there is an entry of `type` `Quantum Espresso DFT SinglePoint` associated with `quartz-scf.out` showing the same errors as before
  * there is a second entry associated with `quartz-scf.out` that has no `type` assigned and only the `data/metadata` section filled. No errors appear in the `logs` section
  * the entries associated with `quartz-nmr.out` and `quartz-efg.out` are the same as in the previous cases.
  
* Files included in the upload: `quartz-scf.out` + `quartz-nmr-gipaw.xml` or `quartz-efg-gipaw.xml`. The behavior is the same as if only `quartz-scf.out` had been uploaded. I assume that the parser for the GIPAW XML file format has not been implemented yet.

ladinesa:
1-2. This is using still electronicparsers.quantum_espresso parser, remove the plugin from your dev env.
3. The error is caused by workflow normalizer not able to build the tasks since the outputs are not present. 
4. The parser for xml outputs is already implemented but these are secondary mainfiles meaning they are only parsed if the corresponding `.out` files are present in the same folder. This is done in order to prevent duplicate entries.

cecizag:
> 1-2. This is using still electronicparsers.quantum_espresso parser, remove the plugin from your dev env.

The previous tests were done by removing the line `"nomad-parser-plugins-electronic>=1.0"` from the dependencies in `pyproject.toml`, but I noticed that `electronicparsers:quantumespresso_parser_entry_point` still appears in the parser list under `about/information`.

So, I kept the dependency list unchanged and excluded the parser by modifying the `nomad.yaml` as follows:

```yaml
keycloak:
  realm_name: "fairdi_nomad_test"
plugins:
  exclude:
  - "electronicparsers:quantumespresso_parser_entry_point"
```

Even in this case, `electronicparsers:quantumespresso_parser_entry_point` still shows up in the parser list under `about/information`, but the test results on the uploads change as follows:

* Files uploaded: only `quartz-scf.out`. Everything works fine as before.

* Files uploaded: `quartz-scf.out` + `quartz-nmr.out`. Only one of the two entries named `quartz-scf.out` in the first screenshot below contains the `data/data` section.
image
image

* Files uploaded: only `quartz.xml`. Process status is `SUCCESS`, but there are these errors in the logs:
image
image

* Files uploaded: `quartz.xml` + `quartz-nmr-gipaw.xml`. Only the entry corresponding to `quartz.xml` is present, the same as in the previous point.

What should I do to completely exclude `electronicparsers:quantumespresso_parser_entry_point`?


> 3. The error is caused by workflow normalizer not able to build the tasks since the outputs are not present.

What do you mean by "the outputs are not present"? The file `quartz-scf.out` is present.


> 4. The parser for xml outputs is already implemented but these are secondary mainfiles meaning they are only parsed if the corresponding `.out` files are present in the same folder. This is done in order to prevent duplicate entries.

I don’t fully understand the logic for parsing the auxiliary XML files. I will ask you for more details during Monday’s meeting.

ladinesa:
1. you may want to do uv sync or even to uv remove <plugin> to update your installations.
2. I can look at the problem with only `quartz.xml` uploaded. I think this is again caused by `model_system` not being parsed which results in the normalizer problem in `nomad_simulations` which should also be fixed.
3. I suspect there are no outputs section created under data. Can you browse the archive and look at `data.outputs` to verify this?
4. Sorry it should have been: "parsed if the corresponding .out files are NOT present in the same folder" meaning only the xml files are uploaded.

cecizag:
> 1. you may want to do uv sync or even to uv remove  to update your installations.

Even after running `uv sync`, `electronicparsers:quantumespresso_parser_entry_point` still appears in the parser list under about/information.



> 3\. I suspect there are no outputs section created under data. Can you browse the archive and look at `data.outputs` to verify this?

After updating the `pyproject.toml` to exclude `electronicparsers:quantumespresso_parser_entry_point`, the situation described in the first comment can no longer be reproduced.



> 4\. Sorry it should have been: "parsed if the corresponding .out files are NOT present in the same folder" meaning only the xml files are uploaded.

Then something must be wrong, because no entry is created for `quartz-nmr-gipaw.xml`, whether the corresponding SCF is in `.out` or `.xml` format. In both cases, there are no other `.out` files for the GIPAW NMR part:
image
image


In addition, I had forgotten to perform this test:

Files in the upload: `quartz-scf.out` + `quartz-nmr.out` + `quartz-efg.out`.
Here is the result: in both entries with `process status` set to `SUCCESS`, the following errors appear in the log and the `data.outputs` section is empty:

image
image

It seems strange to me, especially because it looks like the opposite behavior of what happens when uploading `quartz-scf.out` + `quartz-nmr.out`, as shown in the previous comment.

ladinesa:
For the issue that the `quartz-nmr-gipaw.xml` not being parsed, I made a mistake in the parser spec resulting in the scf xml being the only one matched. I just pushed a fix to resolve this.

cecizag:
I pulled the latest changes and reprocessed the uploads:

* `quartz-scf.out` + `quartz-nmr-gipaw.xml`: there is still only the entry for `quartz-scf.out`

* `quartz.xml` + `quartz-nmr-gipaw.xml`: now there are 2 entries, and the one for `quartz-nmr-gipaw.xml` shows the following errors in the logs:

image
image
image

ladinesa:
1. The xml files cannot be parsed if the .out files are present. The parser is constrained to match only either scf or xml files but not both. There is no possibility to parse both even if we say that they belong to different entries. The way around this is to put them in different folders.
2. I expect three entries similar to the .out case, I will push a fix for this. The other way being for the generic workflow that connects the scf and gipaw entry.

ladinesa:
I added a fix for the workflow.

cecizag:
> I added a fix for the workflow.

* Files in the upload: `quartz-scf.out` + `quartz-nmr.out` or `quartz.xml` + `quartz-nmr-gipaw.xml`: three entries are correctly created—one for SCF, one for NMR, and one for the workflow. The NMR entry still shows the following error:

  ```
  "root":{
      "normalizer":string"MetainfoNormalizer"
      "event":string"Incorrect number of tasks found."
      "proc":string"Entry"
      "process":string"_process_entry"
      "process_worker_id":string"thxu65D3SVaVPhpe4YaGXg"
      "parser":string"nomad_simulation_parsers.parsers:quantumespresso_parser"
      "step":string"MetainfoNormalizer"
      "logger":string"nomad.processing"
      "timestamp":string"2025-10-02T11:40:25Z"
      "level":string"ERROR"
  }
  ```

* Files in the upload: `quartz-scf.out` + `quartz-nmr.out` + `quartz-efg.out`: the workflow entry is missing. The same error as before remains in the NMR and EFG entries.

image

ladinesa:
The `Incorrect number of tasks` error we can ignore for now as this is due to the fact that the `data.outputs` section are not generated as no output properties are currently parsed for the gipaw outputs. I will have a look at the scf-nmr-efg. Just to give me an idea about this workflow, are they supposed to be run in serial in the order scf-nmr-efg? I actually look at the execution time of the calculations to determine the order.

ladinesa:
The output files that I have at the moment have these execution times: `quartz-scf.out`: `starts on 26Mar2025 at 12:12:37`, `quartz-nmr.out`: `starts on 26Mar2025 at 12:13:24`, `quartz-efg.out`: `starts on 26Mar2025 at 11:41:44`. The parser expects the scf output to be executed first and if it is not the case it would not create a workflow. Am i understanding it wrong?

cecizag:
The timestamps I have uploaded are the following:

* scf: `Program PWSCF v.6.4.1 starts on 29May2019 at 13:28:12`
* nmr: `Program GIPAW v.6.4.1 starts on 29May2019 at 13:28:49`
* efg: `Program GIPAW v.6.4.1 starts on 29May2019 at 13:34:05`

In any case, when we discussed with Andra Albino a criterion for linking the GIPAW outputs with those of the self-consistent calculation, the timestamps were never mentioned. We simply agreed that:

* If for one SCF file there is only one NMR file and/or one EFG file, the corresponding GIPAW entries are created and linked to the SCF via the workflow.
* If for one SCF file there is more than one NMR or EFG file, the GIPAW entries are not created and a warning is shown in the logs.
* If the upload contains more than one SCF file along with NMR or EFG files, the GIPAW entries are not created and a warning is shown in the logs.
* To upload multiple SCF files with their corresponding GIPAW files in the same upload, it is necessary to provide a `.zip` archive where each SCF–GIPAW pairing is made explicit by placing the corresponding files in the same subfolder.

Honestly, I can’t say whether the timestamp constraints are correct. We should ask the scientists who run this type of simulation.

ladinesa:
Can you provide me the files? I can easily also lift the timestamp constraints no problem. I am actually able to produce the workflow archive if i adjust the execution time accordingly.

image


One more thing can you disable `temporal` I find it somehow unreliable but you do not have to worry it is a separate issue we will fix.  add this to your `nomad.yaml` file: 
```
temporal:
  enabled: false
```

cecizag:
here you can find the files:

https://github.com/dceresoli/qe-gipaw/tree/7.4/examples/quartz

I think that we can decide if it is better to lift the timestamp constraints on monday.

cecizag:
Here’s a to-do list of the items discussed in today’s meeting:

- [x] Remove the timestamp constraint, or keep it but include a warning in the logs explaining why the uploaded GIPAW files were not parsed.

- [x] Check whether it is correct that, in the upload screen, the `Type` column appears empty.

- [x] Ensure compatibility between SCF and GIPAW entries in different formats. For example, the following combinations should be valid:

  * SCF in XML format and NMR in text format
  * SCF in text format and NMR in XML format

- [x] For clarity, here’s a summary of the version-handling logic I mentioned:

  During testing, a bug was identified in QE-GIPAW that caused an error in the XML output related to the `magnetic_susceptibilities` field.
  This issue was fixed starting from version `7.4.1`.
  Therefore, version handling has been implemented as follows:

  * If the version is `≥ 7.4.1`, the priority is: XML, then text.
  * If the version is `< 7.4.1`:

    * If only the XML file is available, the field is not parsed.
    * If the text file is also available, the field is parsed from the text file instead.

ladinesa:
3. The primary mainfile for QE is .out, so this is partially taken cared out. I cannot change to xml since most of our QE entries only have .out files. I can perhaps make a workaround only for gipaw to parse xml instead of out for v>7.4.1. But it would still show that the mainfile is .out but the parser in the background parser the xml file.

JFRudzinski (approved):

PR #128, opened 2025-11-21
Fix usage of representation schema in mdparserutils

Bernadette-Mohr:
In `mdparserutils.MDParser`, using `Representation` directly raises:
```
nomad-distro-dev/packages/nomad-FAIR/nomad/metainfo/metainfo.py:1353: SyntaxWarning: Representation(lattice_vectors, periodic_boundary_conditions) is not derived from its definition nomad_simulations.schema_packages.model_system.ModelSystem.representations:SubSection.
```
- Updated import to use AlternativeRepresentation from model_system
- Changed parse_trajectory_step parameter type
- Changed instantiation in parse_trajectory_step
- Resolves SyntaxWarning about schema validation

### DIFF
--- a/src/nomad_simulation_parsers/parsers/utils/mdparserutils.py
+++ b/src/nomad_simulation_parsers/parsers/utils/mdparserutils.py
@@ -29,7 +29,6 @@ from nomad_simulations.schema_packages import workflow
 from nomad_simulations.schema_packages.atoms_state import ParticleState
 from nomad_simulations.schema_packages.general import Simulation
 from nomad_simulations.schema_packages.model_system import (
-    AlternativeRepresentation,
     ModelSystem,
 )
 
@@ -148,7 +147,7 @@ class MDParser(ArchiveWriter):
         data: dict[str, Any],
         simulation: Simulation,
         model_system: ModelSystem = None,
-        representation: AlternativeRepresentation = None,
+        # representations: AlternativeRepresentation = None,
     ) -> None:
         """
         Create a system section and write the provided data.
@@ -160,8 +159,6 @@ class MDParser(ArchiveWriter):
             return
         if model_system is None:
             model_system = ModelSystem()
-        if representation is None:
-            representation = AlternativeRepresentation()
 
         lattice_vectors = data.pop('lattice_vectors')
         periodic_boundary_conditions = data.pop('periodic_boundary_conditions')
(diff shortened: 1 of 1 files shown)


src/nomad_simulation_parsers/parsers/utils/mdparserutils.py
```diff
@@ -225,7 +238,7 @@ def parse_md_workflow(self, data: dict[str, Any]) -> None:
         if self.archive is None:
             return
 
-        sec_workflow = MolecularDynamics()
+        sec_workflow = workflow.MolecularDynamics()
```
ndaelman-hu, line 235: this line could remain the same if you made the import more specific

src/nomad_simulation_parsers/parsers/utils/mdparserutils.py
```diff
@@ -158,15 +161,25 @@ def parse_trajectory_step(
...
+        dimensions = data.pop('dimensions')
+
         for label in particle_labels:
             particle_state = ParticleState(label=label)
             model_system.particle_states.append(particle_state)
-        self.parse_section(cell_dict, representation)
-        model_system.representations.append(representation)
+
+        q_lattice = model_system.m_get_quantity_definition('lattice_vectors')
+        model_system.m_set(q_lattice, lattice_vectors)
```
ndaelman-hu: Interesting way of setting quantity values. I typically set them directly, the `m_def` magic  will ensure compliance. Is there any benefit to this approach?

ndaelman-hu: Scanning over the infrastructure, they appear more like legacy.

src/nomad_simulation_parsers/parsers/utils/mdparserutils.py
```diff
@@ -158,15 +161,25 @@ def parse_trajectory_step(
         if model_system is None:
             model_system = ModelSystem()
         if representation is None:
-            representation = Representation()
+            representation = AlternativeRepresentation()
```
ndaelman-hu: When is this is ever used? Note that the target would be `model_system.representations` (plural), which is a repeating subsection.

ndaelman-hu (commented):
This is technically correct. Just not sure that the purpose of `representation` here is.

src/nomad_simulation_parsers/parsers/utils/mdparserutils.py
Bernadette-Mohr: I did run into an issue with directly setting the quantities, which is why I found the complicated way. But since I can't reproduce it anymore, and the direct approach seems to work now, I'm perfectly fine with that.

src/nomad_simulation_parsers/parsers/utils/mdparserutils.py
Bernadette-Mohr: I had left it from the first implementation of `Representation`, maybe we should discuss if the opportunity to set box quantities via representations should be an option? I don't know enough about the intended use cases to make this decision.

ndaelman-hu: What are box quantities? Descriptors of the simulation cell?
`AlternativeRepresentation` is solely for defining different boxes, coordinates, local symmetries, etc. wrt the original representation which falls directly under `ModelMethod`.

ndaelman-hu (approved):
All suggestions were applied

PR #131, opened 2025-12-16
Mapping guide doc

EBB2675:
here is a PR to work together on a practical guide on mapping annotation


ndaelman-hu:
I will just for now give a review solely on the docs. Can do one later the rest of the parser, if necessary.

EBB2675:
> I will just for now give a review solely on the docs. Can do one later the rest of the parser, if necessary.

yes please :handshake: 

i have just quickly pushed the doc from a random branch of mine, the rest is indeed not relevant 
@ndaelman-hu @JFRudzinski @Bernadette-Mohr @kubanmar

docs/mapping-annotations-for-AreaC.md
```diff
@@ -0,0 +1,200 @@
+# Practical Guide to Mapping Annotations for Parser dev
+
+This doc explains how to populate the NOMAD simulations datamodel using mapping annotations. It shows what they are, how to write them, and how to handle nested and repeating sections. 
+
+---
+
+## What a mapping annotation is
+
+- A small note on a schema field that says: *for source `<KEY>`, read value at `<PATH>` and put it here.*
```
ndaelman-hu, line 9:
```suggestion
- A small note on a schema field that says: *for source `<KEY>`, read value at `<PATH>` and place it in the archive under the path leading to the annotated schema field.*
```

docs/mapping-annotations-for-AreaC.md
```diff
@@ -0,0 +1,200 @@
...
+- Stored on any section/quantity: `m_annotations['mapping'][<KEY>] = MapperAnnotation(mapper=...)`.
+- `<KEY>` names the source (e.g., `OUT_KEY`, `HDF5_KEY`, `INFO_KEY`, `BAND_KEY`).
```
ndaelman-hu, line 11: I would add "why we there may be multiple keys". If there was only one, that field would not be necessary.

ndaelman-hu, line 11: Minor nitpick: it doesn't "name" the source per se (as in it does not define the source file).
It just defines a particular source format / mapping.

docs/mapping-annotations-for-AreaC.md
```diff
@@ -0,0 +1,200 @@
...
+- Paths use JMESPath style. Relative paths start with `.` (from the parent’s mapped node). `@` anchors at the source root.
```
ndaelman-hu, line 12:
```suggestion
- Paths use [JMESPath](https://jmespath.org/) style. Relative paths start with `.` (from the parent’s mapped node). `@` anchors at the source root.
```

docs/mapping-annotations-for-AreaC.md
```diff
@@ -0,0 +1,200 @@
...
+- The machinery (`MetainfoParser`) walks the schema, follows these notes, instantiates sections/subsections, and fills values.
+
+## What happens under the hood
+
+1) A source parser (`XMLParser`, `HDF5Parser`, `TextParser`, etc.) reads a file into a Python dict.
+2) Set `annotation_key` on a `MetainfoParser` to the `<KEY>` that matches the annotations.
+3) Call `convert(source_parser, target_parser)`. One key → one `convert`. Multiple keys/sources → run `convert` once per key after switching `annotation_key` each time.
```
ndaelman-hu, line 19: > One key → one `convert`.

Lol, sounds oddly religious.

docs/mapping-annotations-for-AreaC.md
```diff
@@ -0,0 +1,200 @@
...
+4) The target archive is populated; repeating subsections are created automatically from lists.
+
+## Advantages of using mapping
+
+- Declarative and readable: data movement is described alongside the schema, not scattered in code.
+- Reusable across sources: same schema, different annotation keys; swap sources by switching `annotation_key`.
+- Automatic instantiation: nested and repeating subsections are created based on list structure; no manual wiring.
```
ndaelman-hu, line 26: Could you clarify what you mean by "nested" here? It's a very loaded term given the ES component.

I'm asking, as there is a special edge case, but I don't get the feeling that is what you are aiming at here. I simply read it as "a subsection" here, i.e.


```suggestion
- Automatic instantiation of repeating sections and subsections; no manual wiring.
```

docs/mapping-annotations-for-AreaC.md
```diff
@@ -0,0 +1,200 @@
...
+- Composable transforms: reshape, filter, and unit application via small helper functions in the parser.
```
ndaelman-hu, line 27: Agreed. Since you mentioned declarative style above, and these applications are (conceptually) immutable, you could even tentatively call it "functional" here. Jives really well with composition too.

docs/mapping-annotations-for-AreaC.md
```diff
@@ -0,0 +1,200 @@
...
+- Easier maintenance: when input formats change, update paths in one place (annotations) instead of parser code.
+- Parallel branches: ingest multiple files/sources into the same archive by running `convert` per key.
+
+## Path cheat sheet
+
+- Relative from parent: `.field`, `.field[0]`, `.list[*].value`
+- Absolute from root: `field.subfield` (no leading dot)
+- Root anchor: `@`
+- Filter: `.items[?"@name"=="version"] | [0].__value`
+- Attributes (XML/HDF5): keys often prefixed with `@`; values with `__value`
```
ndaelman-hu, line 37: This needs an example to distinguish from the root anchor.

docs/mapping-annotations-for-AreaC.md
```diff
@@ -0,0 +1,200 @@
...
+- Declarative and readable: data movement is described alongside the schema, not scattered in code.
+- Reusable across sources: same schema, different annotation keys; swap sources by switching `annotation_key`.
+- Automatic instantiation: nested and repeating subsections are created based on list structure; no manual wiring.
+- Composable transforms: reshape, filter, and unit application via small helper functions in the parser.
+- Easier maintenance: when input formats change, update paths in one place (annotations) instead of parser code.
+- Parallel branches: ingest multiple files/sources into the same archive by running `convert` per key.
+
+## Path cheat sheet
+
+- Relative from parent: `.field`, `.field[0]`, `.list[*].value`
```
ndaelman-hu, line 33: but no index formula, unfortunately

docs/mapping-annotations-for-AreaC.md
```diff
@@ -0,0 +1,200 @@
...
+)
+```
+
+
+### 7) Units or subfield search
+You can pass `unit='eV'` or `search='.path.to.value'` in the annotation (see Wannier90 eigenvalue/DOS paths).
+
+## Starter template
+
+File: `packages/nomad_simulation_parsers/schema_packages/template.py`
```
ndaelman-hu, line 105: What is the purpose of this line? Regardless, given the size of this template, I would move it out to exactly that file path. You should of course reference it from these docs.

docs/mapping-annotations-for-AreaC.md
```diff
@@ -0,0 +1,200 @@
...
+
+# Transformation example on secondary source
+add_mapping_annotation(
+    outputs.TotalForce.value,
+    SECONDARY_KEY,
+    ('reshape_forces', ['.raw_forces'], {'unit': 'eV / angstrom'}),
+)
+```
+
+File: `packages/nomad_simulation_parsers/parsers/template/parser.py`
```
ndaelman-hu, line 145: Same comment as on the schema-side

docs/mapping-annotations-for-AreaC.md
```diff
@@ -0,0 +1,200 @@
...
+2) Choose a source key (`OUT_KEY`, `HDF5_KEY`, etc.).
+3) Write the JMESPath that matches your source dict. Use `.` to start from the parent node.
+4) Add the annotation with `add_mapping_annotation(target, KEY, path_or_tuple)`.
+5) Repeats: if the subsection is `repeats=True`, just point to a list path; one instance per list element.
+6) If combining multiple sources, run `convert` for each key (set `annotation_key` before calling).
+7) Test on a small file, inspect the archive (e.g., print `archive.data`), and adjust paths.
+
+## Pitfalls to avoid
+
+- Forgetting the leading `.` on relative paths (they become absolute and miss).
```
ndaelman-hu, line 197: So, I don't think this gets covered anywhere, but you raise a valid point:
`.` for relative, `@` for from the root, but what if there is no special character at the front? Where does that path start from?

> Forgetting the leading `.`

I would put the answer higher up. Then this pitfall point can be dropped.

docs/mapping-annotations-for-AreaC.md
```diff
@@ -0,0 +1,200 @@
...
+- Abstract subsections: put annotations on concrete inheriting sections or rely on inheriting annotations.
```
ndaelman-hu, line 198: > Abstract subsections

"Abstract subsections" are those inheriting from `ABC`. Upon re-reading, I guess you mean the imparitive verb mode. If so:


```suggestion
- Avoid annotating every subsection in the chain: put annotations on concrete inheriting sections or rely on inheriting annotations.
```

Still feels like this could become its own section.

docs/mapping-annotations-for-AreaC.md
```diff
@@ -0,0 +1,200 @@
...
+- Reusing a schema with different sources: clear old mappings first with `remove_mapping_annotations`.
```
ndaelman-hu, line 199: When exactly should `remove_mapping_annotations` be called?

docs/mapping-annotations-for-AreaC.md
```diff
@@ -0,0 +1,200 @@
...
+- Relative from parent: `.field`, `.field[0]`, `.list[*].value`
+- Absolute from root: `field.subfield` (no leading dot)
+- Root anchor: `@`
+- Filter: `.items[?"@name"=="version"] | [0].__value`
+- Attributes (XML/HDF5): keys often prefixed with `@`; values with `__value`
+
+## When to annotate `m_def` vs. the attribute
+
+- Attribute-level (on the quantity or subsection attribute):
+  - Use for a specific, local mapping. This is the default for quantities. Example: `Program.version` with `.program_version`.
```
ndaelman-hu, line 42: Local mapping means that it isn't inherited from, correct? I.e. "standalone use"

ndaelman-hu, line 42: Maybe clarify that `Program.version` is the schema, while `.program_version` is the extracted path.

docs/mapping-annotations-for-AreaC.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Core patterns 
+
+### 1) Root anchor
+Anchor the target root section to the **source tree root** (`@` = top of source dict/XML/HDF5).
+```python
+add_mapping_annotation(general.Simulation.m_def, OUT_KEY, '@')
+```
+
+### 2) Nested quantities
```
ndaelman-hu, line 58:
```suggestion
### 2) Quantities
```

Quantities can never be at the root.

docs/mapping-annotations-for-AreaC.md
```diff
@@ -0,0 +1,200 @@
...
+Relative to the parent section’s mapped node.
+```python
+add_mapping_annotation(general.Program.name, OUT_KEY, '.program')
+add_mapping_annotation(general.Program.version, OUT_KEY, '.program_version')
+```
+
+### 3) Single nested subsection
```
ndaelman-hu, line 65:
```suggestion
### 3) Single subsection
```

Tautology: the prefix "sub" already clarifies that is "nested" as you refer to it. Moreover, "nested" obtains a very different meaning when talking about serialization and querying. Would avoid it to prevent confusion.

docs/mapping-annotations-for-AreaC.md
```diff
@@ -0,0 +1,200 @@
...
+Annotate the subsection’s section definition.
+```python
+add_mapping_annotation(model_method.DFT.m_def, OUT_KEY, '.dft')
+```
+
+### 4) Repeating subsections (lists)
```
ndaelman-hu, line 71:
```suggestion
### 4) Repeating subsections & Array-like Quantities
```

docs/mapping-annotations-for-AreaC.md
```diff
@@ -0,0 +1,200 @@
...
+```
+
+### 5) Multiple sources into the same schema
+Use different keys; run `convert` once per key.
+```python
+# Octopus uses OUT_KEY + INFO_KEY + EIGENVALUES_KEY on the same schema
+# Exciting uses INFO_KEY, INPUT_XML_KEY, EIGVAL_KEY, BANDSTRUCTURE_XML_KEY, DOS_XML_KEY
+# Wannier90 uses WOUT_KEY, WIN_KEY, BAND_KEY, WHR_KEY, DOS_KEY
+```
+Set `annotation_key = <KEY>` before each `convert`.
```
ndaelman-hu, line 87: Can you add an example, pls?
From memory, I'd say that `convert` is called at the parser-level instead of the annotations, but might be fuzzy on that one.

docs/mapping-annotations-for-AreaC.md
```diff
@@ -0,0 +1,200 @@
...
+
+### 6) Transformations (reshape, filter, apply units)
+Use a tuple `(function_name, [paths], {kwargs})` and implement the function on the parser.
```
ndaelman-hu, line 90:
```suggestion
Use a tuple `(function_name, [paths], {kwargs})` and implement the function at the parser level.
```

Maybe you should introduce the 2 levels.

ndaelman-hu (commented):
Thank you for adding this. Most of the feedback below comes down to specific phrases.

More importantly, how should we combine this with [the docs in `nomad-docs`](https://github.com/FAIRmat-NFDI/nomad-parser-plugins-simulation/pull/132/changes)?
I would keep both in the same repo, don't have strong feelings which one.

docs/mapping-annotations-for-AreaC.md
EBB2675: this is what i initially wanted to avoid actually, as this internal document is trying to have a casual and simple approach

docs/mapping-annotations-for-AreaC.md
EBB2675: this part is focusing on what, and not why. we should avoid giving out too many details at once

EBB2675: will revise it :+1:

docs/mapping-annotations-for-AreaC.md
EBB2675: it is a brainstorm that we might have example template files stored in the parsers repo to guide people

docs/mapping-annotations-for-AreaC.md
Bernadette-Mohr, line 37: Does "often" here mean that this is a convention, but it would theoretically also be possible, to, e.g., prefix a value with `@`?

Bernadette-Mohr (approved):

docs/mapping-annotations-for-AreaC.md
ndaelman-hu: sure, then maybe add the "why" above

docs/mapping-annotations-for-AreaC.md
ndaelman-hu: Sure. The template is good. I would just move it its own file already.

JFRudzinski:
I will tackle this as part of my documentation work: https://github.com/nomad-coe/nomad-simulations/issues/311

PR #133, opened 2025-12-18
remove annotations fixes

JFRudzinski:
# VASP OUTCAR Parser Fix

## Symptoms
- `tests/parsers/test_vasp_parser.py::test_outcar` passed when executed alone but failed (`AssertionError: No model_method in simulation`) once the full parser suite ran.

## Root Cause
1. The VASP schema lacked mapping entries for the parent `general.Simulation.model_method` subsection, so the OUTCAR mapping could not materialize the parent before filling `model_method.DFT`.
2. Other parser tests (`exciting`, `FHI-aims`, `H5MD`) called `remove_mapping_annotations` with no filter, which wiped *all* mapping keys from the shared `general.Simulation` definition after their runs, erasing the VASP mappings even after they were added.

## Fix
- Added the missing parent annotations in `src/nomad_simulation_parsers/schema_packages/vasp.py`:
  ```python
  add_mapping_annotation(general.Simulation.model_method, XML_KEY, '.parameters')
  add_mapping_annotation(general.Simulation.model_method, OUTCAR_KEY, '.parameters')
  ```
- Extended `remove_mapping_annotations` to accept an `annotation_keys` filter and updated the `exciting`, `FHI-aims`, and `H5MD` parsers to drop only their own mapping keys. This keeps their cleanup behavior without stripping VASP’s `xml`/`outcar` mappings.

## Verification
```
$ uv run --directory packages/nomad-parser-plugins-simulation pytest \
    tests/parsers/test_vasp_parser.py tests/parsers/test_quantumespresso_parser.py
$ uv run --directory packages/nomad-parser-plugins-simulation pytest tests
```
Both commands now pass; `test_outcar` succeeds regardless of suite order.

## Notes
- QE failures (`IndexError: pop index out of range` inside `Simulation.model_system` removal) were ultimately due to the shared mapper cache leaking data between parsers. The [FIX](https://gitlab.mpcdf.mpg.de/nomad-lab/nomad-FAIR/-/merge_requests/2823) lives in `nomad-FAIR`; without it you would have to run QE tests in isolation because there’s no robust way to prevent the cache from reusing ABINIT’s data mid-suite.

### DIFF
--- a/src/nomad_simulation_parsers/parsers/exciting/parser.py
+++ b/src/nomad_simulation_parsers/parsers/exciting/parser.py
@@ -249,7 +249,16 @@ class ExcitingArchiveWriter(ArchiveWriter):
         data_parser.close()
 
         # remove annotations
-        remove_mapping_annotations(exciting.general.Simulation.m_def)
+        remove_mapping_annotations(
+            exciting.general.Simulation.m_def,
+            annotation_keys=[
+                exciting.INFO_KEY,
+                exciting.INPUT_XML_KEY,
+                exciting.EIGVAL_KEY,
+                exciting.BANDSTRUCTURE_XML_KEY,
+                exciting.DOS_XML_KEY,
+            ],
+        )
 
 
 class ExcitingParser(MatchingParser):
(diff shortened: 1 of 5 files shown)


JFRudzinski:
@ndaelman-hu what I don't know is if the solution in nomad-FAIR is only arising because of this fix and/or if there is a different way to solve that issue directly in the parser

PR #137, opened 2026-01-07
Fix global mapping annotation contamination in parsers

ndaelman-hu:
## Purpose

Fixes critical global state contamination issue where mapping annotations were being removed, causing parser failures when multiple parsers run in the same process. This affected VASP tests (required subprocess workaround) and could cause production failures in sequential parsing scenarios.

## Scope

**Included**

- Remove `reload()` calls from all 10 parsers
- Remove `remove_mapping_annotations()` calls from Exciting, FHI-aims, and H5MD parsers
- Delete `remove_mapping_annotations()` utility function  
- Revert VASP subprocess isolation workaround to original simple tests
- Clean up unused imports

**Out of scope**

- Fixing other potential global state issues in NOMAD
- Architectural refactoring of mapping parser system

## Context & Links

Related to test isolation issue documented in `VASP_TEST_ISOLATION_FIX.md`. Root cause: mapping annotations stored on class-level `m_def` objects are shared globally. Two antipatterns caused contamination:

1. `reload(schema_module)` creates new class instances, but NOMAD plugin system references original `m_def`
2. `remove_mapping_annotations()` removes annotations from shared global state

Annotations should persist as static metadata for process lifetime, not be dynamically added/removed per parse.

## Reviewer Notes

All parsers now follow VASP's pattern of static annotations. Test carefully that:
- Sequential parsing of different file types works in same process
- Long-running worker processes don't accumulate issues
- No parser-specific cleanup logic was needed for legitimate reasons

## Status

- [x] Ready for review

## Breaking Changes

- [ ] None

## Dependencies / Blockers

- [ ] None

## Testing / Validation

- [x] Unit tests - All 42 parser tests pass in both isolation and full suite
- [x] Manual testing - Verified annotations persist correctly across multiple parses

**Before:** VASP test_outcar failed in full suite, required subprocess isolation
**After:** All tests pass, no workarounds needed

### DIFF
--- a/src/nomad_simulation_parsers/parsers/__init__.py
+++ b/src/nomad_simulation_parsers/parsers/__init__.py
@@ -191,7 +191,7 @@ vasp_parser = EntryPoint(
     mainfile_contents_re=(
         r'^\s*<\?xml version="1\.0" encoding="ISO-8859-1"\?>\s*?\s*<modeling>?\s*'
         r'<generator>?\s*<i name="program" type="string">\s*vasp\s*</i>?|'
-        r'^\svasp[\.\d]+.+?(?:\(build|complex)[\s\S]+?executed on'
+        r'^\s*vasp[\.\d]+.+?(?:\(build|complex)[\s\S]+?executed on'
     ),
     mainfile_mime_re='(application/.*)|(text/.*)',
     mainfile_name_re='.*[^/]*xml[^/]*',
--- a/src/nomad_simulation_parsers/parsers/__init__.py
+++ b/src/nomad_simulation_parsers/parsers/__init__.py
@@ -25,6 +25,7 @@ class EntryPoint(ParserEntryPoint):
             LOGGER.error(
                 f'Could not load parser class {self.parser_class_name}', exc_info=e
             )
+            raise
 
 
 class Wannier90ParserEntryPoint(EntryPoint):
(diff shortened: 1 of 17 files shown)


ladinesa:
We really need a reloading mechanism that works  with metainfo. This is necessary as the annotation can be overwritten by other parsers.

ndaelman-hu:
> We really need a reloading mechanism that works with metainfo. This is necessary as the annotation can be overwritten by other parsers.

I think that's a completely valid point. I want to leave the design to your discretion to implement it as you want, so I just put this PR up as a suggested solution without touching your code.

The issue (to my understanding) is that the annotations are read in all at once, and then fully wiped after the first parser finishes.[1] My feature suggestion is to provide a list of the keys that should be removed. An empty list then either flushes nothing or everything.

[1] Bit more elaboration: firstly, I verified this and pushed an example to the branch `showcase-mappingannotation-removal-bug`. You can try running the whole test set locally, it'll fail in 5 parsers. This bug was never found before because of a confluence of 2 reasons:

1. Only 4 parsers on the current `develop` ever remove annotations: exciting, fhiaims, h5md, and vasp.
2. None ever tested whether the `data.method` is filled, until my [PR for pseudopotentials](https://github.com/FAIRmat-NFDI/nomad-parser-plugins-simulation/pull/129).

PR #145, opened 2026-02-06
Linting fix

ladinesa:


src/nomad_simulation_parsers/parsers/gpaw/gpw_parser.py
```diff
@@ -234,24 +234,24 @@ def get_array(self, key: str) -> np.ndarray:
         if self.ulm is None:
             return
         values = {
-            'unitcell': lambda: self.ulm.atoms.cell,
```
ndaelman-hu, line 237: typically the dict type is specified before for the dict as a whole, not its individual elements...
good that it was fixed

ndaelman-hu (approved):
Corrected the arguments to `self.split` and a type hint.

JFRudzinski (approved):

PR #154, opened 2026-02-24
Fix QE parser error

EBB2675:
here i resolve ruff PLC0207 by limiting split operations to one split each in QE parser

lets make the CI green again!!!

### DIFF
--- a/src/nomad_simulation_parsers/parsers/quantumespresso/parser.py
+++ b/src/nomad_simulation_parsers/parsers/quantumespresso/parser.py
@@ -140,7 +140,8 @@ class MainfileTextParser(TextParser):
         ]
 
     def get_xc_functionals(self, source: str) -> list[dict[str, Any]]:
-        numbers = source.split('(', maxsplit=1)[1].split(')', maxsplit=1)[0]
+        # Keep legacy behavior for additional '(' while satisfying PLC0207.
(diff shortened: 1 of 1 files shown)


src/nomad_simulation_parsers/parsers/quantumespresso/parser.py
```diff
@@ -140,7 +140,7 @@ def get_energy_contributions(self, source: dict[str, Any]):
         ]
 
     def get_xc_functionals(self, source: str) -> list[dict[str, Any]]:
-        numbers = source.split('(')[1].split(')')[0]
+        numbers = source.split('(', maxsplit=1)[1].split(')', maxsplit=1)[0]
         nval = (4, 10)
```
Bernadette-Mohr, line 145: If that still gets the right results, great!

Bernadette-Mohr (approved):

src/nomad_simulation_parsers/parsers/quantumespresso/parser.py
EBB2675: good point, i tightened it a bit more :+1: 

`split('(', maxsplit=2)[1]` is behaviorally equivalent to `split('(')[1]`, 
and `split(')', maxsplit=1)[0]` is equivalent to `split(')')[0]`

PR #158, opened 2026-03-17
Extend yambo spectra

Jajar26:
Work in progress to share

l275-304 
 #start HB
    def get_spectra(self) -> dict[str, Any]:
        data = []
        names = []

        with open(self.filepath) as f:
            for line in f:
                line = line.strip()

                if line.startswith('#') and 'E/ev' in line:
                    names = [k.strip() for k in line.split() if k != '#']
                    continue

                if names and not line.startswith('#') and line:
                    data.append(line.split())

        if not data:
            return {}

        data = np.array(data, dtype=np.float64)

        if data.shape[1] < 2:
            return {}

        return dict(
            excitation_energies=data[:, 0] * ureg.eV,
            intensities=data[:, 1]
        )

        #end HB

l345-373

 #start HB
        SPECTRA_TYPE_MAP = {
        'Absorption': 'dielectric_function',
        'EELS': 'energy_loss_spectrum',
        }

        for spectra_file in spectra_files:
            spectra_parser.filepath = spectra_file

            sp_type = mainfile_parser.data.get('sp_type')
            if sp_type is None:
                continue

            absorption_spectra_parser.data_object = yambo.outputs.AbsorptionSpectrum()

            spectra_parser.convert(absorption_spectra_parser)

            spectra_obj = absorption_spectra_parser.data_object
            spectra_obj.label = sp_type
            spectra_obj.type = SPECTRA_TYPE_MAP.get(sp_type, 'unknown')

            outputs = (
                data.outputs[-1] if data.outputs else data.m_create(yambo.outputs.Outputs)
            )
            outputs.m_append(
                yambo.outputs.AbsorptionSpectrum.m_def,
                spectra_obj,
            )
            #end HB


src/nomad_simulation_parsers/parsers/yambo/parser.py
```diff
@@ -272,8 +272,36 @@ class YamboSpectraParser(TextParser):
...
         return LOGGER
 
-    def get_spectra(self, data: np.ndarray) -> dict[str, Any]:
-        return dict(excitation_energies=data[:, 0], intensities=data[:, 1])
+    #start HB
+    def get_spectra(self) -> dict[str, Any]:
+        data = []
+        names = []
+
+        with open(self.filepath) as f:
```
ladinesa, line 286: we should introduce a new file parser class for the spectra file, we do not want to append the data to the mainfile parser.

src/nomad_simulation_parsers/parsers/yambo/parser.py
```diff
@@ -313,19 +341,34 @@ def write_to_archive(self):
...
+
+            sp_type = mainfile_parser.data.get('sp_type')
+            if sp_type is None:
+                continue
+
             spectra_parser.convert(absorption_spectra_parser)
-            # add to simulation outputs.absorption_spectra
+
+            spectra_obj = absorption_spectra_parser.data_object
+            spectra_obj.label = sp_type
```
ladinesa: we should not do this here but instead add this in the annotation

src/nomad_simulation_parsers/parsers/yambo/parser.py
```diff
@@ -313,19 +341,34 @@ def write_to_archive(self):
...
+            spectra_obj.type = SPECTRA_TYPE_MAP.get(sp_type, 'unknown')
+
             outputs = (
-                data.outputs[-1]
-                if data.outputs
-                else data.m_create(yambo.outputs.Outputs)
+                data.outputs[-1] if data.outputs else data.m_create(yambo.outputs.Outputs)
             )
             outputs.m_append(
```
ladinesa, line 396: we should also add this to the Outputs annotation inorder to automate this

src/nomad_simulation_parsers/parsers/yambo/parser.py
```diff
@@ -313,19 +341,34 @@ def write_to_archive(self):
         absorption_spectra_parser = YamboMetainfoParser()
         absorption_spectra_parser.data_object = yambo.outputs.AbsorptionSpectrum()
         absorption_spectra_parser.annotation_key = yambo.SPECTRA_KEY
+        
+        #start HB
+        SPECTRA_TYPE_MAP = {
```
ladinesa, line 368: add this to init

src/nomad_simulation_parsers/parsers/yambo/parser.py
emolteni: Sorry, maybe I'm missing something....  By "introducing a new file parser class for the spectra file" do you mean adding something to the definition of the class SpectraParser in file_parsers.py?  Apart from that, in parser.py we indeed have a  class YamboSpectraParser (into which my colleague Hajar and I have added some lines..).   Moreover, are we really appending spectra data to the mainfile parser?

ladinesa: Yes indeed this is what I meant. We should have a file parser  and a mapping parser. What you are trying to do at the moment is parsing the spectra file and appending it to a function call within the mainfile parser. Although this could work, I would like to as much as possible separate the auxiliary files.

emolteni: Ok, so, if I understand correctly, what you suggest may mean handling spectra files similarly to how netcdf files are already handled, i.e. we should   1) "populate" the SpectraParser class definition in file_parser.py, possibly putting there some parts that currently are in the YamboSpectraParser class in parser.py,    2)  making the YamboSpectraParser class (in parser.py) of  MappingParser type, instead of TestParser type.

ladinesa: Yes indeed. You should put the part that search for spectra files in `YamboArchiveWriter`.  You then create a `YamboSpectraParser` (mapping parser) in the archive writer similar to the net cdf case. You should add the corresponding annotations in the schema file and use a different annotation key e.g. `yambo_spectra`

ladinesa:
Can you please add some test data with the spectra files?

src/nomad_simulation_parsers/parsers/yambo/parser.py
```diff
@@ -267,13 +267,43 @@ def unpack_modules(modules: list[dict[str, Any]]) -> list[dict[str, Any]]:
...
+    self.spectra_parser = SpectraParser(mainfile=self.filepath)   # EM
+    
     @property
     def logger(self):
         return LOGGER
 
-    def get_spectra(self, data: np.ndarray) -> dict[str, Any]:
-        return dict(excitation_energies=data[:, 0], intensities=data[:, 1])
+    #start HB
+    def get_spectra(self) -> dict[str, Any]:
```
ladinesa, line 280: this is already covered by the file parser I believe. DataTextParser will simply use np.loadtxt to read the file.

src/nomad_simulation_parsers/parsers/yambo/parser.py
```diff
@@ -313,19 +343,34 @@ def write_to_archive(self):
...
+        #start HB
+        SPECTRA_TYPE_MAP = {
+        'Absorption': 'dielectric_function',
+        'EELS': 'energy_loss_spectrum',
+        }
+
         for spectra_file in spectra_files:
             spectra_parser.filepath = spectra_file
+
+            sp_type = mainfile_parser.data.get('sp_type')
```
ladinesa: this should be spectra_parser.get('sp_type') as you defined it there.

src/nomad_simulation_parsers/schema_packages/yambo.py
```diff
@@ -67,13 +67,21 @@ class AbsorptionSpectra(outputs.AbsorptionSpectrum):
     add_mapping_annotation(
         outputs.AbsorptionSpectrum.energies, SPECTRA_KEY, '.excitation_energies'
     )
+# EM, 6 Jul, 2026:    
+    add_mapping_annotation(
+        outputs.AbsorptionSpectrum.sp_type, SPECTRA_KEY, 'sp_type'
+    )
+    add_mapping_annotation(
+        outputs.AbsorptionSpectrum.n_energies, SPECTRA_KEY, 'n_energies'
```
ladinesa: n_energies is not defined in the spectra parser.

src/nomad_simulation_parsers/schema_packages/yambo.py
```diff
@@ -67,13 +67,21 @@ class AbsorptionSpectra(outputs.AbsorptionSpectrum):
...
+    )
+# end EM
 
 
 class Outputs(outputs.Outputs):
     # TODO add description
-    sp_type = Quantity(type=str)
+#    sp_type = Quantity(type=str)   # EM: commented,  6 Jul, 2026
```
ladinesa, line 101: is `sp_type` already in the nomad-simulations Outputs?

emolteni: Indeed it isn't.  So, here in /schema_packages/yambo.py, should I uncomment it in the class Outputs or rather put it in the class AbsorptionSpectra?

src/nomad_simulation_parsers/parsers/yambo/parser.py
emolteni: Done

src/nomad_simulation_parsers/schema_packages/yambo.py
ladinesa: Ah i get it now, sorry it is already in AbsorptionSpectra.

ladinesa:
With the added test data, can you please also add the tests for the yambo parser?

emolteni:
> With the added test data, can you please also add the tests for the yambo parser?

I have just created a tentative version of a test_yambo_parser.py file for testing spectra parsing, trying to follow the structure of other test parsers in that folder. I am not so confident about it.....

As for my /test/data/yambo folder, instead, I had to upload its /SAVE folder zipped due to file size limits, and for the same reason the system does not allow me to unzip it there on github.  The SAVE folder (which contains the ns.db file) should not be needed for spectra parsing, but its absence is expected to lead to missing parsing of atom coordinates, chemical composition, cell parameters, which in "production" parsing is unpleasant.  If you think we can ignore this for the tests, then I can just delete the /SAVE folder.

src/nomad_simulation_parsers/schema_packages/yambo.py
ladinesa: adsorption_spectra is not in nomad-simulation Outputs. so maybe add
```
 adsorption_spectra = SubSection(sub_section=AbsorptionSpectra.m_def, repeats=True)
```
under class Outputs in yambo.py

tests/parsers/test_yambo_parser.py
```diff
@@ -0,0 +1,14 @@
...
+
+from nomad_simulation_parsers.parsers.yambo  import YamboArchiveWriter   
+
+LOGGER = get_logger(__name__)
+
+def test_spectra(parser):
+  archive = EntryArchive()
+  parser = YamboArchiveWriter()
+  parser.parse('tests/data/yambo/r_setup',archive,LOGGER)
+  assert spectra_files[0] = 'o-R_methylox_TDLDA.alpha_q1_slepc_alda_bse'
```
ladinesa, line 12: The assertion here should be 
```
spectra = archive.data.outputs[0].adsorption_spectra[0]
assert spectra.sp_type == 'Polarizability'
...
```

src/nomad_simulation_parsers/schema_packages/yambo.py
ladinesa: you should also add an annotation for Outputs.m_def here  with SPECTRA_KEY something like

```
add_mapping_annotation(Outputs.m_def, SPECTRA_KEY, '.@')
```

ladinesa, line 151: also for general.Simulation.m_def

tests/parsers/test_yambo_parser.py
emolteni: Done, i.e. I have put your lines in place of my previous line on checking spectra type.  

I expect the previous line, which should check it finds our spectra file, is ok? There I have spectra_files[0] because, in principle, spectra_file is a list, but in the present case it contains 1 file only.

ladinesa: there is no `spectra_file` defined under `AdsorptionSpectra` only: `value`, `energies`, `sp_type`and `n_energies`

src/nomad_simulation_parsers/parsers/yambo/parser.py
```diff
@@ -313,19 +343,34 @@ def write_to_archive(self):
...
+            sp_type = spectra_parser.get('sp_type')  # EM: spectra_parser.get,  Jul 8, 2026 
+            if sp_type is None:
+                continue
+
             spectra_parser.convert(absorption_spectra_parser)
-            # add to simulation outputs.absorption_spectra
+
+            spectra_obj = absorption_spectra_parser.data_object
+            spectra_obj.label = sp_type
+            spectra_obj.type = SPECTRA_TYPE_MAP.get(sp_type, 'unknown')
```
ladinesa: also, `label`and `type` are not defined in `AdsorptionSpectra`

src/nomad_simulation_parsers/schema_packages/yambo.py
emolteni: Well, in fact we do have a line: 
absorption_spectra = SubSection(sub_section=AbsorptionSpectrum.m_def, repeats=True)

in  nomad_simulations/schema_packages/outputs.py. 


But it seems we never use this "absorption_spectra" anywhere, am I wrong?   I'll see if I can understand this while I fix the other issues you mentioned.

ladinesa: I do not see it here: https://github.com/FAIRmat-NFDI/nomad-simulations/blob/develop/src/nomad_simulations/schema_packages/outputs.py

src/nomad_simulation_parsers/schema_packages/yambo.py
emolteni: Done

emolteni: Done (line 125), but for some reason it does not label it as "outdated" here.

src/nomad_simulation_parsers/schema_packages/yambo.py
emolteni: Well, I see it there, at line 148.

ladinesa: my bad, I was searching for `adsorption_spectra`

emolteni: Ok, perfect.  About the misspell, I had noticed it in your messages but forgot to point it out.  In this context, it is always absorption.

src/nomad_simulation_parsers/schema_packages/yambo.py
ladinesa: since we would like to use the new definition for `AbsorptionSpectrum`, this should be: `AbsorptionSpectrum.m_def` not `outputs.AbsorptionSpectrum.m_def`

src/nomad_simulation_parsers/schema_packages/yambo.py
emolteni: Yes, indeed. I wish to define it in the SpectraParser class in yambo/file_parsers.py, but in fact I am not sure how to define it, since it is a derived quantity (it is shape[0] of   data.outputs, or anyway of the object that contains all the (energy,intensity) pairs of values of a spectrum), not defined by matching a specific regular expression (at a difference with sp_type and many other quantities in that or other file_parsers.py).  

There might be an alternative way of defining n_energies, by looking for the line containing the expression "Energy steps" in the o* file(s).

Jajar26:
Hi,

I made the following changes:

* fixed the regex pattern used to extract "n_energies"  in "file_parsers.py" ;
* refactored the "AbsorptionSpectra" class to use direct inheritance from "AbsorptionSpectrum" instead of "outputs.AbsorptionSpectrum" and updated the related mapping accordingly;
* tried to fix things that didn't seem logical to me.

I also opened a separate PR on "nomad-simulations" (#450 https://github.com/FAIRmat-NFDI/nomad-simulations/pull/450/changes/bc5680002b020eeb499258497266061ab2ba9bcc) to add the "sp_type" and "n_energies" quantities to "AbsorptionSpectrum".

I do not have access to the NOMAD Oasis environment, so I could not test the complete workflow there. I tested the changes as much as possible on my side and I hope I did not introduce any issues. Please let me know if anything needs to be adjusted.

emolteni:
Hi, 
Hajar and I have made some further progress (as shown in the above commits and comments) since your last review of our PR.

Mainly: 
- created the Quantity n_energies in file_parser.py using a regex, i.e. by directly fetching the numerical value of that quantity (labeled either BEnSteps or ETStpsXd in the spectrum file) instead of counting the spectrum lines in that file; n_energies is then used in parser.py
- opened a separated PR   (#450 https://github.com/FAIRmat-NFDI/nomad-simulations/commit/bc5680002b020eeb499258497266061ab2ba9bcc)  for adding  the "sp_type" and "n_energies" quantities to "AbsorptionSpectrum" in nomad_simulations/schema_packages/properties/spectral_profile.py 
- refactored the "AbsorptionSpectra" class to use direct inheritance from "AbsorptionSpectrum" instead of "outputs.AbsorptionSpectrum" and updated the related mapping accordingly
- .. and further small changes.

So I think that when you have time you can have a look and let us know your comments/ further suggestions. 
Thanks again

ladinesa:
> Hi, Hajar and I have made some further progress (as shown in the above commits and comments) since your last review of our PR.
> 
> Mainly:
> 
>     * created the Quantity n_energies in file_parser.py using a regex, i.e. by directly fetching the numerical value of that quantity (labeled either BEnSteps or ETStpsXd in the spectrum file) instead of counting the spectrum lines in that file; n_energies is then used in parser.py
> 
>     * opened a separated PR   (#450 [FAIRmat-NFDI/nomad-simulations@bc56800](https://github.com/FAIRmat-NFDI/nomad-simulations/commit/bc5680002b020eeb499258497266061ab2ba9bcc))  for adding  the "sp_type" and "n_energies" quantities to "AbsorptionSpectrum" in nomad_simulations/schema_packages/properties/spectral_profile.py
> 
>     * refactored the "AbsorptionSpectra" class to use direct inheritance from "AbsorptionSpectrum" instead of "outputs.AbsorptionSpectrum" and updated the related mapping accordingly
> 
>     * .. and further small changes.
> 
> 
> So I think that when you have time you can have a look and let us know your comments/ further suggestions. Thanks again

I can have look this week.

PR #176, opened 2026-05-18
Backup/test data normalization fhiaims 2026 05 12

JFRudzinski:
fhi-aims only changes from #170


JFRudzinski:
@ndaelman-hu Do you know about this? Can you check if this support is already covered in the fhiaims parser?

ndaelman-hu:
Closing this backup branch. Its FHI-aims carve-out is the commented-out `get_band_gaps` transformer, which is not covered on develop: the `electronic_band_gaps` annotation (`schema_packages/fhiaims.py:187`) references a `get_band_gaps` method that does not exist, so band gaps are never populated for FHI-aims. This is not a #170 regression (baseline never mapped `humo`/`lumo` into `archive.data`), so it is tracked separately in #206 rather than revived here.

PR #177, opened 2026-05-27
Add FHI-aims SCF convergence criteria extraction

ndaelman-hu:
**Note**: This PR supersedes #173, which was closed after history rewrite to clean up accidentally committed files.

---

## Purpose

Fixes critical data loss bug in multi-mapper pipeline when parsing polymorphic subsections. Previously, only 1 out of N sibling mappers would successfully populate data.

## Scope

**Included**
- Test fix: Update `test_scf_convergence_criteria` expectations to match actual convergence values from test data
- Test fix: Update `test_k_mesh` to handle polymorphic `numerical_settings` list
- Ruff fixes: Line length, unused imports, f-string issues

**Out of scope**
- Core fix is in nomad-FAIR (see Context & Links)

## Context & Links

**Depends on**: nomad-lab/nomad-FAIR!3040

This PR contains test updates and code cleanup for the FHI-aims parser. The core bug fix is in the nomad-FAIR mapping parser and must be merged first.

## Reviewer Notes

The `install-and-test` CI job will fail until nomad-FAIR!3040 is merged and released to stable, because the tests depend on the mapping parser fix. The `ruff` and `pytest` jobs validate that the code changes are correct.

## Status

- [x] Ready for review
- [x] Blocked: Waiting for nomad-FAIR!3040 to merge

## Breaking Changes

- [x] None

## Dependencies / Blockers

- [x] External dependencies: Requires nomad-FAIR!3002

## Testing / Validation

- [x] Unit tests: Updated test expectations to match actual data
- [x] Manual testing: Verified parsed archive preserves all 4 numerical_settings items (previously only 1)

### DIFF
--- a/src/nomad_simulation_parsers/parsers/fhiaims/parser.py
+++ b/src/nomad_simulation_parsers/parsers/fhiaims/parser.py
@@ -319,61 +319,71 @@ class FHIAimsOutMappingParser(TextMappingParser):
             return np.array([0.0, 0.0, 0.0])
         return k_offset
 
-    def get_scf_convergence_criteria(
-        self, source: dict[str, Any]
-    ) -> list[dict[str, Any]]:
+    def get_scf_energy_criterion(self, source: dict[str, Any]) -> dict[str, Any]:
         """
-        Extract SCF convergence criteria and return as list of SelfConsistency dicts.
-
-        FHI-aims supports multiple convergence criteria:
-        - Total energy convergence (eV)
-        - Charge density convergence (dimensionless)
-        - Sum of eigenvalues convergence (eV)
-        - Maximum number of SCF iterations
+        Extract energy convergence criterion for SCF.
 
         Returns:
-            List of dictionaries, each representing one SelfConsistency section.
+            Dictionary representing energy SelfConsistency section, or empty dict.
         """
-        convergence_criteria = []
-
-        # Energy convergence threshold
         conv_energy = source.get('convergence_energy')
-        if conv_energy is not None:
-            convergence_criteria.append(
-                {
-                    'threshold_change': conv_energy,
-                    'threshold_change_unit': 'energy',
-                }
-            )
+        if conv_energy is None:
+            return {}
 
-        # Charge density convergence threshold
+        result = {
+            'threshold_change': conv_energy,
+            'threshold_change_unit': 'energy',
+        }
+
+        max_iterations = source.get('max_scf_iterations')
+        if max_iterations is not None:
+            result['n_max_iterations'] = max_iterations
+
+        return result
+
+    def get_scf_density_criterion(self, source: dict[str, Any]) -> dict[str, Any]:
+        """
+        Extract density convergence criterion for SCF.
+
+        Returns:
+            Dictionary representing density SelfConsistency section, or empty dict.
+        """
         conv_density = source.get('convergence_density')
-        if conv_density is not None:
-            convergence_criteria.append(
-                {
-                    'threshold_change': conv_density,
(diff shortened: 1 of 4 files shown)


tests/parsers/test_fhiaims_parser.py
```diff
@@ -124,3 +125,50 @@ def test_k_mesh(tmp_path, k_offset_line, expected_offset):
...
+    assert density_criterion is not None
+    assert density_criterion.threshold_change == approx(1.0e-5)
+
+    # Check energy and eigenvalues convergence criteria (both in eV)
+    # Need to distinguish by threshold value since both use 'eV'
+    ev_criteria = [sc for sc in scf_criteria if sc.threshold_change_unit == 'eV']
+    assert len(ev_criteria) == 2
+
+    # Energy criterion has smaller threshold (1e-6 eV)
+    energy_criterion = next(sc for sc in ev_criteria if sc.threshold_change < 1.0e-5)
```
EBB2675: From what I understood, the total-energy and sum-of-eigenvalues criteria both appear as `SelfConsistency` entries with `threshold_change_unit='eV'`, so the archive can only distinguish them by the numerical threshold value. This test also seems to rely on that distinction by magnitude, which feels semantically fragile. Would it make sense to preserve the criterion target in `SelfConsistency.name`, or through another explicit field/convention?

EBB2675 (approved):
thanks alot, looks good from my side

I just wrote one minor question about how we semantically distinguish the different `SelfConsistency` instances in the archive, especially when multiple criteria use the same unit

tests/parsers/test_fhiaims_parser.py
ndaelman-hu: Thank you for the suggestion. Added the `name` field now.

src/nomad_simulation_parsers/parsers/fhiaims/parser.py
```diff
@@ -319,6 +319,75 @@ def get_k_offset_with_default(self, k_offset: np.ndarray | None) -> np.ndarray:
...
+            Dictionary representing energy SelfConsistency section, or empty dict.
+        """
+        conv_energy = source.get('convergence_energy')
+        if conv_energy is None:
+            return {}
+
+        result = {
+            'name': 'total_energy_change',
+            'threshold_change': conv_energy,
+            'threshold_change_unit': 'eV',
```
ladinesa: annotations take also unit as aeg

ndaelman-hu: I agree, this would be better, but then I have to update the current schema (adding `flexible_unit=True`). Should work, but lemme check first.

ndaelman-hu: Updated the schema at https://github.com/FAIRmat-NFDI/nomad-simulations/pull/393

ladinesa:
I managed to include the self consistency section to numerical settings, please see #181

src/nomad_simulation_parsers/parsers/fhiaims/parser.py
```diff
@@ -511,6 +578,42 @@ def write_to_archive(  # noqa: PLR0915
 
         out_parser.convert(archive_handler, remove=False)
 
+        # Manually create SelfConsistency instances for SCF convergence criteria
+        # (multi-mapper doesn't support function-returns-list pattern)
+        from nomad_simulations.schema_packages.numerical_settings import (  # noqa: PLC0415
+            SelfConsistency,
+        )
+
+        if self.archive.data.model_method:
```
ladinesa, line 587: I think you can simply put this in the annotation see my pr.

PR #181, opened 2026-06-01
Add scf annotation

ladinesa:
Remove the manual creation of scf numerical setting section.

### DIFF
--- a/src/nomad_simulation_parsers/parsers/fhiaims/parser.py
+++ b/src/nomad_simulation_parsers/parsers/fhiaims/parser.py
@@ -267,18 +267,18 @@ class FHIAimsOutMappingParser(TextMappingParser):
             return np.array([0.0, 0.0, 0.0])
         return k_offset
 
-    def get_all_criteria(self) -> list[dict[str, Any]]:
+    def get_all_criteria(self, source: dict[str, Any]) -> list[dict[str, Any]]:
         criteria = []
         for fcriteria in [
             self.get_scf_energy_criterion,
             self.get_scf_density_criterion,
             self.get_scf_eigenvalues_criterion,
         ]:
-            data = fcriteria(self.data)
+            data = fcriteria(source)
             if data:
                 criteria.append(data)
         if not criteria:
-            max_iterations = self.data.get('max_scf_iterations')
+            max_iterations = source.get('max_scf_iterations')
             if max_iterations is not None:
                 criteria.append({'n_max_iterations': max_iterations})
         return criteria
@@ -588,7 +588,7 @@ class FHIAimsArchiveWriter(ArchiveWriter):
             if energy_threshold is not None:
                 self.archive.workflow2.method.convergence_targets = [
                     EnergyConvergenceTarget(
-                        threshold=energy_threshold * ureg.eV,
+                        threshold=energy_threshold,
                         threshold_type='absolute',
                     )
                 ]
--- a/src/nomad_simulation_parsers/parsers/fhiaims/parser.py
+++ b/src/nomad_simulation_parsers/parsers/fhiaims/parser.py
@@ -278,14 +278,6 @@ class FHIAimsOutMappingParser(TextMappingParser):
         if conv_energy is None:
             return {}
 
-        # Extract magnitude and unit from pint Quantity if present
-        if hasattr(conv_energy, 'magnitude') and hasattr(conv_energy, 'units'):
-            value = conv_energy.magnitude
-            unit = format(conv_energy.units, '~P')  # Use short symbol format
-        else:
-            value = conv_energy
-            unit = 'eV'
-
         result = {
             'name': 'total_energy_change',
             'threshold_change': conv_energy,
@@ -330,16 +322,6 @@ class FHIAimsOutMappingParser(TextMappingParser):
         conv_eigenvalues = source.get('convergence_eigenvalues')
         if conv_eigenvalues is None:
             return {}
-
-        # Extract magnitude and unit from pint Quantity if present
-        if hasattr(conv_eigenvalues, 'magnitude') and hasattr(
-            conv_eigenvalues, 'units'
-        ):
-            value = conv_eigenvalues.magnitude
-            unit = format(conv_eigenvalues.units, '~P')  # Use short symbol format
-        else:
-            value = conv_eigenvalues
-            unit = 'eV'
 
         result = {
             'name': 'sum_eigenvalues_change',
--- a/src/nomad_simulation_parsers/parsers/fhiaims/parser.py
+++ b/src/nomad_simulation_parsers/parsers/fhiaims/parser.py
@@ -278,6 +278,14 @@ class FHIAimsOutMappingParser(TextMappingParser):
         if conv_energy is None:
             return {}
 
+        # Extract magnitude and unit from pint Quantity if present
+        if hasattr(conv_energy, 'magnitude') and hasattr(conv_energy, 'units'):
+            value = conv_energy.magnitude
+            unit = format(conv_energy.units, '~P')  # Use short symbol format
+        else:
+            value = conv_energy
+            unit = 'eV'
+
         result = {
             'name': 'total_energy_change',
             'threshold_change': conv_energy,
@@ -322,6 +330,16 @@ class FHIAimsOutMappingParser(TextMappingParser):
         conv_eigenvalues = source.get('convergence_eigenvalues')
         if conv_eigenvalues is None:
             return {}
+
+        # Extract magnitude and unit from pint Quantity if present
+        if hasattr(conv_eigenvalues, 'magnitude') and hasattr(
+            conv_eigenvalues, 'units'
+        ):
+            value = conv_eigenvalues.magnitude
+            unit = format(conv_eigenvalues.units, '~P')  # Use short symbol format
+        else:
+            value = conv_eigenvalues
+            unit = 'eV'
 
         result = {
             'name': 'sum_eigenvalues_change',
--- a/src/nomad_simulation_parsers/parsers/fhiaims/parser.py
+++ b/src/nomad_simulation_parsers/parsers/fhiaims/parser.py
@@ -267,6 +267,10 @@ class FHIAimsOutMappingParser(TextMappingParser):
             return np.array([0.0, 0.0, 0.0])
         return k_offset
 
+    def get_all_criteria(self) -> list[dict[str, Any]]:
+        criteria = [self.get_scf_energy_criterion(self.data)]
+        return criteria
+
     def get_scf_energy_criterion(self, source: dict[str, Any]) -> dict[str, Any]:
         """
         Extract energy convergence criterion for SCF.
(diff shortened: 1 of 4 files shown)


ndaelman-hu (changes requested):
I'm sorry, but this implementation breaks the tests for me locally.

Moreover, parsing the `Si_geomopt/out.out` test file, only generates a single criterion (energy), whereas it should extract 3: energy, density, and eigenvalues.
Some fields (e.g. `SelfConsistency.name` or `n_max_iterations`) are either incorrectly filled, or left blank. This is minor compared to the criteria issue, though.

ladinesa:
> I'm sorry, but this implementation breaks the tests for me locally.
> 
> Moreover, parsing the `Si_geomopt/out.out` test file, only generates a single criterion (energy), whereas it should extract 3: energy, density, and eigenvalues. Some fields (e.g. `SelfConsistency.name` or `n_max_iterations`) are either incorrectly filled, or left blank. This is minor compared to the criteria issue, though.

This is not complete implementation, it is just to give you an idea how to improve it

ladinesa:
@ndaelman-hu @EBB2675 @JFRudzinski 

my initial idea to simplify the fhiaims parsing of scf conv quantites. Feel free to improve on it further.

ndaelman-hu:
> @ndaelman-hu @EBB2675 @JFRudzinski
> 
> my initial idea to simplify the fhiaims parsing of scf conv quantites. Feel free to improve on it further.

Thx! I'll try parsing again with this to see whether it extracts all criteria.

ladinesa:
@ndaelman-hu  should we simply merge this?

ladinesa:
@ndaelman-hu can we merge this?

ndaelman-hu:
> @ndaelman-hu can we merge this?

Thank you for reminding me, @ladinesa . Just ran the parser on the test files, and I get the following error for `csae_2/aims.out`:

```python
File ".../parsers/fhiaims/parser.py", line 781, in parse
File ".../parsers/fhiaims/parser.py", line 643, in write_to_archive
    threshold=energy_threshold * ureg.eV,
  File ".../pint/facets/plain/quantity.py", line 989, in _mul_div
    if not self._check(other):
  File ".../pint/util.py", line 991, in _check
    raise ValueError(
ValueError: Cannot operate with Quantity and Unit of different registries.
```

I would remove the additional unit in `threshold=energy_threshold * ureg.eV`.
Either you or me can apply this. Am also double-checking the code statically, will soon be done.

src/nomad_simulation_parsers/schema_packages/fhiaims.py
```diff
@@ -88,8 +88,25 @@ class Program(general.Program):
     add_mapping_annotation(general.Program.version, TEXT_KEY, '.version')
 
 
+class SelfConsistency(numerical_settings.SelfConsistency):
+    add_mapping_annotation(
+        numerical_settings.SelfConsistency.threshold_change,
+        TEXT_KEY,
+        '.threshold_change',
+    )
+    add_mapping_annotation(numerical_settings.SelfConsistency.name, TEXT_KEY, '.name')
```
ndaelman-hu, line 100: **`n_max_iterations` is dropped.** Only `.threshold_change`/`.name` are mapped, but the criterion dicts and the no-threshold fallback carry `n_max_iterations`, so it never reaches the archive and the fallback yields an empty section. Regression vs the removed `SelfConsistency(**criterion)`.
```suggestion
    add_mapping_annotation(numerical_settings.SelfConsistency.name, TEXT_KEY, '.name')
    add_mapping_annotation(
        numerical_settings.SelfConsistency.n_max_iterations,
        TEXT_KEY,
        '.n_max_iterations',
    )
```

src/nomad_simulation_parsers/parsers/fhiaims/parser.py
```diff
@@ -319,6 +319,22 @@ def get_k_offset_with_default(self, k_offset: np.ndarray | None) -> np.ndarray:
...
+            self.get_scf_eigenvalues_criterion,
+        ]:
+            data = fcriteria(self.data)
+            if data:
+                criteria.append(data)
+        if not criteria:
+            max_iterations = self.data.get('max_scf_iterations')
+            if max_iterations is not None:
+                criteria.append({'n_max_iterations': max_iterations})
+        return criteria
```
ndaelman-hu: **Reaches into `self.data`.** Convention is to receive the source via the annotation, not read `self.data`. The `get_scf_*_criterion` helpers already take a `source`, so just thread one through here and pair with `('get_all_criteria', ['.@'])` on the schema side.
```suggestion
    def get_all_criteria(self, source: dict[str, Any]) -> list[dict[str, Any]]:
        criteria = []
        for fcriteria in [
            self.get_scf_energy_criterion,
            self.get_scf_density_criterion,
            self.get_scf_eigenvalues_criterion,
        ]:
            data = fcriteria(source)
            if data:
                criteria.append(data)
        if not criteria:
            max_iterations = source.get('max_scf_iterations')
            if max_iterations is not None:
                criteria.append({'n_max_iterations': max_iterations})
        return criteria
```

src/nomad_simulation_parsers/schema_packages/fhiaims.py
```diff
@@ -88,8 +88,25 @@ class Program(general.Program):
...
+
+
 class DFT(model_method.DFT):
-    add_mapping_annotation(numerical_settings.KSpace.m_def, TEXT_KEY, '.@')
+    add_mapping_annotation(
+        numerical_settings.SelfConsistency.m_def,
+        TEXT_KEY,
+        ('get_all_criteria', []),
```
ndaelman-hu: **The empty arg list is what forces the `self.data` read above.** The second tuple element is the positional argument list handed to `get_all_criteria`; `[]` invokes it with no source, so the transformer has no choice but to reach into `self.data`. `['.@']` materializes the current source node and passes it in as the `source` argument, which is what keeps the transformer declarative. Apply this together with the `get_all_criteria(self, source)` suggestion, one without the other breaks the call (the annotation would pass an argument the function does not accept, or vice versa).
```suggestion
        ('get_all_criteria', ['.@']),
```

src/nomad_simulation_parsers/schema_packages/fhiaims.py
```diff
@@ -52,7 +52,7 @@ class Simulation(general.Simulation):
         ),
     )
     # DFT method - only annotate DFT.m_def, not ModelMethod base class
-    add_mapping_annotation(model_method.DFT.m_def, TEXT_KEY, '.@')
+    add_mapping_annotation(model_method.DFT.m_def, TEXT_KEY, '.@', update_mode='merge')
```
ndaelman-hu: **Why `merge` here?** `DFT` is now written by more than one annotation: this base `.@` map, the XC `functional_key` data from the recent functional work, and the `DFT`-class subsections below (`SelfConsistency`, `KSpace`). Under the default mode a later write can replace the section rather than extend it, so the last annotation applied wins and earlier fields are silently lost. `merge` folds the contributions into a single `DFT`. Please confirm that is the intent, and that it genuinely composes with the XC annotations rather than papering over a section collision that would be better resolved at the source.

src/nomad_simulation_parsers/schema_packages/fhiaims.py
```diff
@@ -88,8 +88,25 @@ class Program(general.Program):
...
 class DFT(model_method.DFT):
-    add_mapping_annotation(numerical_settings.KSpace.m_def, TEXT_KEY, '.@')
+    add_mapping_annotation(
+        numerical_settings.SelfConsistency.m_def,
+        TEXT_KEY,
+        ('get_all_criteria', []),
+        update_mode='append',
+    )
+    add_mapping_annotation(
+        numerical_settings.KSpace.m_def, TEXT_KEY, '.@', update_mode='append'
```
ndaelman-hu: **Why `append` for a single-node source?** `append` adds a new `KSpace` for each matched source node. That is the correct mode for the `SelfConsistency.m_def` sibling above, since `get_all_criteria` returns a list and each element becomes its own section, but `KSpace` maps a single `.@`, so append should only ever produce one. Confirm there is exactly one k-space node per entry, otherwise this silently emits duplicate (or empty) `KSpace` sections, and that switching from the previous plain `.@` was deliberate rather than mirrored from the `SelfConsistency` line.

ndaelman-hu (commented):
This PR refactors the SCF section for FHIaims to the match the updated schema. Most of the inline suggestions just push the declarative mapping further and should be confirmed against the intended behavior. One property, `n_max_iterations`, was lost in the refactor.

ladinesa:
> ureg

already removed this in a separate mr,

src/nomad_simulation_parsers/parsers/fhiaims/parser.py
ladinesa: added arg

src/nomad_simulation_parsers/schema_packages/fhiaims.py
ladinesa: added

ladinesa:
> > @ndaelman-hu can we merge this?
> 
> Thank you for reminding me, @ladinesa . Just ran the parser on the test files, and I get the following error for `csae_2/aims.out`:
> 
> ```python
> File ".../parsers/fhiaims/parser.py", line 781, in parse
> File ".../parsers/fhiaims/parser.py", line 643, in write_to_archive
>     threshold=energy_threshold * ureg.eV,
>   File ".../pint/facets/plain/quantity.py", line 989, in _mul_div
>     if not self._check(other):
>   File ".../pint/util.py", line 991, in _check
>     raise ValueError(
> ValueError: Cannot operate with Quantity and Unit of different registries.
> ```
> 
> I would remove the additional unit in `threshold=energy_threshold * ureg.eV`. Either you or me can apply this. Am also double-checking the code statically, will soon be done.

removed

src/nomad_simulation_parsers/schema_packages/fhiaims.py
ladinesa: does not really matter, no other method section aside from DFT, also the default. I can also remove.

src/nomad_simulation_parsers/schema_packages/fhiaims.py
ladinesa: we should not merge it with the self consistency settings right?

src/nomad_simulation_parsers/schema_packages/fhiaims.py
ndaelman-hu: Also applied, I see.

src/nomad_simulation_parsers/schema_packages/fhiaims.py
ndaelman-hu: Thx, I checked and XC, SCF, and KSpace all are still populated.

src/nomad_simulation_parsers/schema_packages/fhiaims.py
ndaelman-hu: That's not quite what I meant. On the current code state, the two `KSpace.m_def` annotations register under the same `TEXT_KEY`, so `add_mapping_annotation`'s `.update()` makes the second (plain `.@`) overwrite the first — `update_mode='append'` never takes effect.

**Verified locally on single-point and geometry-optimization files**: exactly one `KSpace` section either way, so no duplicate/empty sections, but the `append` line is dead code.

Please collapse lines 116–119 to:
```python
add_mapping_annotation(numerical_settings.KSpace.m_def, TEXT_KEY, '.@')
```
Runtime-identical, just drops the misleading `append`.

ndaelman-hu (approved):
This PR replaces the FHI-aims parser's manual `SelfConsistency` construction with a declarative `get_all_criteria` mapping annotation that extracts all three SCF convergence criteria (energy, density, eigenvalues).

My previous review comments have all been addressed, though one could be slightly ameliorated.

Good to merge, but first resolve the current merge conflicts pls. No need for further review.

PR #182, opened 2026-06-02
ORCA parser

EBB2675:
Parse representative single-point ORCA calculations, including program name and version, model systems, DFT, RI-MP2, correlated-method HF references, DLPNO-CCSD(T), CASCI/QD-NEVPT2, selected local-correlation settings, basis sets, scalar relativistic treatments, and canonical molecular orbitals for supported restricted-output layouts. Initial CASSCF parsing and serial local-correlation workflow construction are included but require additional integration coverage.


src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,925 @@
...
+        for mo_index, coefficient in zip(mo_indices, row_values):
+            coefficients_by_mo.setdefault(mo_index, []).append(coefficient)
+
+    return _coefficient_matrix(coefficients_by_mo)
+
+
+def _token_list_to_mo_coefficients(tokens: list[str]) -> np.ndarray | None:
+    coefficients_by_mo: dict[int, list[float]] = {}
+    index = 0
+    while index < len(tokens):
```
ladinesa, line 103: for loop?

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,925 @@
...
+        if not tokens[index].isdigit():
+            index += 1
+            continue
+
+        mo_indices = []
+        while index < len(tokens) and tokens[index].isdigit():
```
ladinesa: also here

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,925 @@
...
+
+
+def _read_coefficient_rows(
+    tokens: list[str],
+    index: int,
+    mo_indices: list[int],
+    coefficients_by_mo: dict[int, list[float]],
+) -> int:
+    n_indices = len(mo_indices)
+    while index < len(tokens) and AO_ROW_RE.fullmatch(tokens[index]):
```
ladinesa: also here

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,925 @@
...
+        return None
+
+    return np.asarray(coefficients, dtype=np.float64)
+
+
+class OutParser(MappingTextParser):
+    def __init__(self) -> None:
+        super().__init__(text_parser=OutReader())
+        self.parse_only_required = False
+        self.text_parser.parse_only_required = False
```
ladinesa: I think text_parser does not have attr parse_only_required.

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,925 @@
...
+        self.text_parser.findlazy = False
+
+    def load_file(self) -> OutReader:
```
ladinesa, line 184: you can do super().load_file() then set findlazy to False to save some lines

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,925 @@
...
+        if self.filepath:
+            self.text_parser.findlazy = False
+            self.text_parser.mainfile = self.filepath
+        return self.text_parser
+
+    @staticmethod
+    def _as_dict(value: Any) -> dict[str, Any]:
```
ladinesa: not sure about this extra handling? TextParser should do it.

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,925 @@
...
+
+    @staticmethod
+    def _scalar(value: Any) -> Any:
+        if isinstance(value, np.ndarray):
+            value = value.tolist()
+        if isinstance(value, (list, tuple)):
+            return value[0] if value else None
+        return value
+
+    def get_program_name(self, src: dict[str, Any]) -> str:
```
ladinesa: you can remove this and the annotation and set directly the name in the program section

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,925 @@
...
+        return 'ORCA'
+
+    def get_atoms(self, src: dict[str, Any]) -> list[dict[str, Any]]:
```
ladinesa: since this returns only one system and does not do any special transformation, i think the reqular path annotation can handle it. for particle states, maybe you can do use function

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,925 @@
...
+        return {
+            'RHF': 'RKS',
+            'UHF': 'UKS',
+            'ROHF': 'ROKS',
+            'RKS': 'RKS',
+            'UKS': 'UKS',
+            'ROKS': 'ROKS',
+        }.get(reference.upper())
+
+    def get_dft(self, src: dict[str, Any]) -> dict[str, Any]:
```
ladinesa: also here, maybe try to split it into smaller functions, so it is more transparent.

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,925 @@
...
+        if not xc:
+            return {}
+
+        method = {'xc': xc}
+        reference_form = self._dft_reference_form(scf_settings.get('hf_type'))
+        if reference_form:
+            method['reference_form'] = reference_form
+        return method
+
+    def _get_multireference_method_data(self, source: dict[str, Any]) -> dict[str, Any]:
```
ladinesa: i would suggest also here.

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,925 @@
...
+            'state_treatment': (
+                'state_averaged' if len(state_weights) > 1 else 'state_specific'
+            ),
+            'n_state_groups': len(state_multiplicities) or None,
+            'state_multiplicities': state_multiplicities or None,
+            'n_roots_per_multiplicity': n_roots_per_multiplicity or None,
+            'state_weights': state_weights or None,
+        }
+
+    def get_multireference_scf_methods(
```
ladinesa, line 402: i think this would be unnecessary as you will not have method if cassf is None?

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,925 @@
...
+        reader = OutParser()
+        reader.filepath = mainfile
+        archive.data = Simulation()
+        metainfo_parser = MetainfoParser(data_object=archive.data)
+        metainfo_parser.annotation_key = orca.OUT_KEY
+        metainfo_parser.max_nested_level = 3
+
+        try:
+            reader.convert(metainfo_parser)
+            reader.enrich_methods(archive.data)
```
ladinesa: We try to manage this with mapping annotations, I can try to open a pr from this.

ladinesa:
As promised, I put up some simplifications in #185

src/nomad_simulation_parsers/parsers/orca/parser.py
EBB2675: because the MO coefficient blocks have variable column and row counts. The next block start can only be determined after consuming the current block, so the loop needs to update the token index dynamically. I added a comment to make this explicit

ndaelman-hu:
MO HOMO-LUMO gap works against both the current `develop` schema and the incoming `MOGap` rework (#471), with no parser changes.

ORCA prints only the `ORBITAL ENERGIES` block (per-orbital energy + occupation) — no explicit gap or HOMO/LUMO line. The `nomad-simulations` schema derives the frontier gap in `MolecularOrbitals.normalize()` (reached via the `MetainfoNormalizer` pass, not `Simulation.normalize()`) from the `value`, `occupations`, and `kind='canonical'` this parser already maps.

Verified end-to-end with `normalize_all`:
- **`develop`:** yields `homo_normalized` / `lumo_normalized` / `homo_lumo_gap_normalized`, self-consistent (`gap == lumo - homo`).
- **#471 (`fix/frontier-gap-resolver`):** yields a `FrontierLevels(is_derived=True)` under `MolecularOrbitals.frontier_levels` and a derived `MOGap` (with `derived_from` set) under `Outputs.molecular_orbital_gaps` — identical values.

So this PR stays on `develop`; adopting `MOGap` is a follow-up that only re-pins `nomad-simulations` and swaps the gap assertion, with no parser code. A parsed (`is_derived=False`) gap would only be emitted if an ORCA output reported one directly, which none of the current fixtures do.

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+    def parse(
+        self,
+        mainfile: str,
+        archive: 'EntryArchive',
+        logger: 'BoundLogger',
+        child_archives: dict[str, 'EntryArchive'] | None = None,
+    ) -> None:
+        reload(orca)
+
+        reader = OutParser()
```
ladinesa: introduce a writer wrapper similar to other parsers.

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+
+class OrcaParser(MatchingParser):
+    def parse(
+        self,
+        mainfile: str,
+        archive: 'EntryArchive',
+        logger: 'BoundLogger',
+        child_archives: dict[str, 'EntryArchive'] | None = None,
+    ) -> None:
+        reload(orca)
```
ladinesa: remove and use orca-unique keys

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+            if data.get('local_correlation')
+            else None,
+            numerical_settings=[
+                cls._build_local_settings(settings)
+                for settings in data.get('numerical_settings', [])
+            ],
+        )
+
+    @classmethod
+    def _build_cc(cls, data: dict[str, Any]) -> CC:
```
ladinesa: this is not used?

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+    def _build_local_settings(data: dict[str, Any]) -> LocalCorrelationSettings:
+        return LocalCorrelationSettings(
+            screening_thresholds=[
+                LocalCorrelationThreshold(**threshold)
+                for threshold in data.get('screening_thresholds', [])
+            ]
+        )
+
+    @classmethod
+    def _build_perturbation(cls, data: dict[str, Any]) -> PerturbationMethod:
```
ladinesa: this is not used?

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+        return LocalCorrelation(
+            type=data.get('type'),
+            spaces=[
+                LocalCorrelationSpace(**space_data)
+                for space_data in data.get('spaces', [])
+            ],
+        )
+
+    @staticmethod
+    def _build_local_settings(data: dict[str, Any]) -> LocalCorrelationSettings:
```
ladinesa: this is not used?

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+                'molecular_orbitals': molecular_orbitals,
+            }
+        ]
+
+    @staticmethod
+    def _build_active_space(data: dict[str, Any] | None) -> ActiveSpace | None:
+        return ActiveSpace(**data) if data else None
+
+    @staticmethod
+    def _build_local_correlation(data: dict[str, Any]) -> LocalCorrelation:
```
ladinesa: this is not used?

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+
+        return [
+            {
+                'model_system_ref': '/data/model_system/0',
+                'molecular_orbitals': molecular_orbitals,
+            }
+        ]
+
+    @staticmethod
+    def _build_active_space(data: dict[str, Any] | None) -> ActiveSpace | None:
```
ladinesa: this is not used?

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+            table = np.asarray(orbital_energies, dtype=np.float64)
+            if (
+                table.ndim != ORBITAL_ENERGY_N_DIMENSIONS
+                or table.shape[1] < ORBITAL_ENERGY_N_COLUMNS
+            ):
+                table = None
+
+        n_ao = self._scalar(basis_set_total.get('main_basis_set'))
+        molecular_orbitals = {
+            'm_def': MolecularOrbitals.m_def.qualified_name(),
```
ladinesa: unnecessary to include m_def, mapping parser should take care of ti

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+        super().__init__(text_parser=OutReader())
+        self._method = None
+
+    def load_file(self) -> OutReader:
+        text_parser = super().load_file()
+        text_parser.findlazy = False
+        return text_parser
+
+    @staticmethod
+    def _parser_results(value: Any) -> dict[str, Any]:
```
ladinesa: why is this wrapper necessary, FileParser should have a get method

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+        )
+        if method not in electronic_methods:
+            return []
+        model = self.get_relativity_model(source)
+        model.setdefault('basis_set', {})
+        return [model]
+
+    @staticmethod
+    def _allowed_basis_set_keys(method: str | None) -> set[str]:
+        if method in ('HF', 'DFT', 'MultireferenceSCF', 'MultireferenceCI'):
```
ladinesa: these supposed to be in init as these are static values

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+            'state_treatment': (
+                'state_averaged' if len(state_weights) > 1 else 'state_specific'
+            ),
+            'n_state_groups': len(state_multiplicities) or None,
+            'state_multiplicities': state_multiplicities or None,
+            'n_roots_per_multiplicity': n_roots_per_multiplicity or None,
+            'state_weights': state_weights or None,
+        }
+
+    def _is_casci(self, source: dict[str, Any], casscf: dict[str, Any]) -> bool:
```
ladinesa: make it local

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+    def _get_casscf_data(self, source: dict[str, Any]) -> dict[str, Any]:
+        return self._navigate(source, 'single_point', 'casscf')
+
+    def get_dft_methods(self, src: dict[str, Any]) -> list[dict[str, Any]]:
+        method = self.get_dft(src)
+        if method:
+            self._method = 'DFT'
+        return [method] if method else []
+
+    def _build_active_space_data(self, casscf: dict[str, Any]) -> dict[str, Any]:
```
ladinesa: make it local

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+        if not xc:
+            return {}
+
+        method = {'xc': xc}
+        reference_form = self._get_dft_reference_form(scf_settings)
+        if reference_form:
+            method['reference_form'] = reference_form
+        return method
+
+    def _get_casscf_data(self, source: dict[str, Any]) -> dict[str, Any]:
```
ladinesa: unnecessary

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+        xc = {}
+        functional_key = self._get_xc_functional_key(scf_settings)
+        if functional_key:
+            xc['functional_key'] = functional_key
+        hf_fraction = self._get_exact_exchange_fraction(scf_settings)
+        if hf_fraction is not None:
+            xc['global_exact_exchange'] = hf_fraction
+        return xc
+
+    def _get_dft_reference_form(self, scf_settings: dict[str, Any]) -> str | None:
```
ladinesa: unnecessary

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+            or scf_settings.get('correl_functional')
+        )
+        functionals = [
+            value
+            for value in (exchange, correlation)
+            if isinstance(value, str) and value
+        ]
+        return '+'.join(dict.fromkeys(functionals))
+
+    def _get_exact_exchange_fraction(self, scf_settings: dict[str, Any]) -> Any:
```
ladinesa: unnecessary

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+        return {
+            'RHF': 'RKS',
+            'UHF': 'UKS',
+            'ROHF': 'ROKS',
+            'RKS': 'RKS',
+            'UKS': 'UKS',
+            'ROKS': 'ROKS',
+        }.get(reference.upper())
+
+    def _get_scf_settings(self, source: dict[str, Any]) -> dict[str, Any]:
```
ladinesa: unnecessary

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+        return self._navigate(source, 'single_point', 'self_consistent', 'scf_settings')
+
+    def _get_xc_functional_key(self, scf_settings: dict[str, Any]) -> str:
```
ladinesa: make local

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+    @staticmethod
+    def _hf_reference_form(value: Any) -> str | None:
+        reference = OutParser._scalar(value)
+        if not isinstance(reference, str):
+            return None
+        reference = reference.upper()
+        return reference if reference in {'RHF', 'UHF', 'ROHF'} else None
+
+    @staticmethod
+    def _dft_reference_form(value: Any) -> str | None:
```
ladinesa: make local

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+            'AIOPM-NBO': 'AIOPM-NBO',
+            'IBO': 'IBO',
+            'NBO': 'NBO',
+        }.items():
+            if key in raw:
+                return normalized
+        return None
+
+    @staticmethod
+    def _hf_reference_form(value: Any) -> str | None:
```
ladinesa: make local

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+            {
+                'is_representative': True,
+                'positions': positions,
+                'particle_states': self._build_particle_states(symbols),
+                **self._get_charge_and_multiplicity(src),
+            }
+        ]
+
+    @staticmethod
+    def _normalize_localization_method(value: Any) -> str | None:
```
ladinesa: make local

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+    def _get_cartesian_system(self, source: dict[str, Any]) -> tuple[list[str], Any]:
+        single_point = self._navigate(source, 'single_point')
+        coordinates = single_point.get('cartesian_coordinates', [])
+        return str_to_cartesian_coordinates(coordinates) if coordinates else ([], None)
+
+    @staticmethod
+    def _build_particle_states(symbols: list[str]) -> list[dict[str, str]]:
+        return [{'chemical_symbol': symbol} for symbol in symbols]
+
+    def _get_charge_and_multiplicity(self, source: dict[str, Any]) -> dict[str, int]:
```
ladinesa: make local

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+            return value[0] if value else None
+        return value
+
+    def _get_cartesian_system(self, source: dict[str, Any]) -> tuple[list[str], Any]:
+        single_point = self._navigate(source, 'single_point')
+        coordinates = single_point.get('cartesian_coordinates', [])
+        return str_to_cartesian_coordinates(coordinates) if coordinates else ([], None)
+
+    @staticmethod
+    def _build_particle_states(symbols: list[str]) -> list[dict[str, str]]:
```
ladinesa: make local

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+
+    @staticmethod
+    def _scalar(value: Any) -> Any:
+        if isinstance(value, np.ndarray):
+            value = value.tolist()
+        if isinstance(value, (list, tuple)):
+            return value[0] if value else None
+        return value
+
+    def _get_cartesian_system(self, source: dict[str, Any]) -> tuple[list[str], Any]:
```
ladinesa: make local

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+        current = source
+        for key in keys:
+            current = self._parser_results(current.get(key))
+        return current
+
+    def _get_mdci_data(self, source: dict[str, Any]) -> dict[str, Any]:
+        return self._navigate(source, 'single_point', 'ci')
+
+    @staticmethod
+    def _scalar(value: Any) -> Any:
```
ladinesa: ~~should not be private~~ rename to _to_scalar, also I would not make it static method if not used somewhere else where context is not present.

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+        # Nested TextParser quantities expose parsed values through `_results`.
+        return value._results if hasattr(value, '_results') else value or {}
+
+    def _navigate(self, source: dict[str, Any], *keys: str) -> dict[str, Any]:
+        current = source
+        for key in keys:
+            current = self._parser_results(current.get(key))
+        return current
+
+    def _get_mdci_data(self, source: dict[str, Any]) -> dict[str, Any]:
```
ladinesa: unnecessary

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+def _read_mo_indices(tokens: list[str], start: int) -> tuple[list[int], int]:
+    mo_indices = []
+    for index, token in enumerate(tokens[start:], start):
+        if not token.isdigit():
+            return mo_indices, index
+        mo_indices.append(int(token))
+    return mo_indices, len(tokens)
+
+
+def _read_coefficient_rows(
```
ladinesa: make local

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+
+
+def _find_next_mo_header(tokens: list[str], start: int) -> int:
+    return next(
+        (index for index, token in enumerate(tokens[start:], start) if token.isdigit()),
+        len(tokens),
+    )
+
+
+def _read_mo_indices(tokens: list[str], start: int) -> tuple[list[int], int]:
```
ladinesa: make local

src/nomad_simulation_parsers/parsers/orca/parser.py
```diff
@@ -0,0 +1,877 @@
...
+        index += n_indices * 2
+        separator_tokens = tokens[index : index + n_indices]
+        index += sum(1 for token in separator_tokens if token.startswith('-'))
+
+        index = _read_coefficient_rows(tokens, index, mo_indices, coefficients_by_mo)
+
+    return _coefficient_matrix(coefficients_by_mo)
+
+
+def _find_next_mo_header(tokens: list[str], start: int) -> int:
```
ladinesa: make local

tests/parsers/test_orca_parser.py
```diff
@@ -0,0 +1,281 @@
+from collections.abc import Generator
```
ladinesa: please adopt new test format so we do not need to migrate

ndaelman-hu:
@ladinesa for `_get_scf_settings` and `_get_casscf_data` (flagged `unnecessary`): each has two call sites, so inlining would duplicate the `_navigate(...)` path — keep them as shared helpers, or still inline?

Otherwise, addressed on this branch:

- **Dead `_build_*` section constructors** (`_build_active_space`, `_build_local_correlation`, `_build_local_settings`, `_build_perturbation`, `_build_cc`) — confirmed unused since direct section creation moved to mapping annotations (#185); removed them and the now-orphaned `model_method` / `numerical_settings` imports.
- **`m_def` injected into the `molecular_orbitals` payload** — dropped; the `MolecularOrbitals.m_def` annotation in `schema_packages/orca.py` already resolves the section (tests still pass without it).

The rest — single-use helpers → local, static key-sets → `__init__`, the `_parser_results` wrapper vs a `FileParser` `get`, the writer wrapper, and orca-unique keys — read as valid; I would take them in a follow-up rather than widen this PR scope.

ladinesa:
My primary concern was there are too many of these private functions that make it a spaghetti. I would remove them. For the _parser_results wrapper, the question was why is this necessary? maybe you can also remove it.

ndaelman-hu:
@ladinesa Agreed — and I think both comments trace to one root cause: the transformers take `['.@']` and then re-walk the fixed path with `_navigate` internally, instead of resolving it in the annotation. The mapping parser already does that descent declaratively (the `.active_space.n_active_electrons` paths cross a nesting level and unwrap fine), so the right fix is to hand each transformer its subtree via a dotted path, e.g.:

```
('get_dft_methods', ['.single_point.self_consistent.scf_settings'])
```

That removes the `_navigate` / `_parser_results` layer and most of the small private helpers in one pass, rather than just inlining them.

This structure is pre-existing — it predates the gap work this PR adds — so rather than widen this PR, I will do the refactor in a **follow-up PR**. I will resolve your remaining comments here first, then tag you for review on the follow-up.

ndaelman-hu (approved):
Adapted the code the current `develop` branch.
As @ladinesa pointed out, previous iterations circumvented the declarative parsing style. This will be amended in a follow-up PR.

ladinesa:
> @ladinesa Agreed — and I think both comments trace to one root cause: the transformers take `['.@']` and then re-walk the fixed path with `_navigate` internally, instead of resolving it in the annotation. The mapping parser already does that descent declaratively (the `.active_space.n_active_electrons` paths cross a nesting level and unwrap fine), so the right fix is to hand each transformer its subtree via a dotted path, e.g.:
> 
> ```
> ('get_dft_methods', ['.single_point.self_consistent.scf_settings'])
> ```
> 
> That removes the `_navigate` / `_parser_results` layer and most of the small private helpers in one pass, rather than just inlining them.
> 
> This structure is pre-existing — it predates the gap work this PR adds — so rather than widen this PR, I will do the refactor in a **follow-up PR**. I will resolve your remaining comments here first, then tag you for review on the follow-up.

ok thanks. then i will approve and please open the new issue(s)

PR #189, opened 2026-07-03
Add syrupy snapshot verification with manually triggered CI job

ndaelman-hu:
## Purpose

Add a permanent, on-demand snapshot regression harness that checks parser output
is **additive**: a source ref must not drop any `archive.data` leaf present in a
target ref. It was built to verify the PR #170 migration (catching regressions
the normal suite missed — GPAW occupations, Octopus `model_system`, VASP
`SelfConsistency` threshold, QE per-step `total_forces`) and is kept as a
general tool for future parser changes.

## Scope

**Included**

- `tests/parsers/test_migration_snapshots.py`: one syrupy snapshot test per parser (15 tests, 14 parsers) serializing `archive.data`
- `tests/parsers/compare_snapshots.py`: structural, index-aligned leaf diff that fails only on **removed** leaves (data loss); tolerates additions and representation changes (e.g. eV→J)
- `tests/parsers/check_additivity.sh`: the generate-both-and-compare orchestration, runnable locally and in CI; takes a `target` (baseline) and `source` (under test) ref, defaulting to `develop` vs `HEAD`
- `.github/workflows/snapshot-tests.yml`: a `workflow_dispatch` wrapper around that script, adding a per-SHA `actions/cache` and artifact upload; inputs `target` (default `develop`) and `source` (default the dispatched branch)
- `syrupy` added to the `dev` extra

Snapshots are **not tracked** — they are generated per run (git-ignored). Each ref's
snapshot is produced in a worktree via a `--no-deps` reinstall so that ref's code
and matching entry points are used while the resolved `nomad-lab` is kept.

**Out of scope**

- The parser corrections themselves (merged with #170)

## Context & Links

- Base is `develop` (PR #170 has merged).
- Requires a `nomad-lab` pre-release from the MPCDF registry (contains nomad-FAIR MR 3103) for deterministic polymorphic sub-section ordering; the repo's primary index is that registry, so CI resolves it. Without it the ordering varies per process and the comparison is flaky.
- The `workflow_dispatch` job cannot be dispatched from Actions until this workflow file is on the default branch (`develop`) — see the comment below. Until then it runs locally via `check_additivity.sh`.

## Reviewer Notes

Run locally: `bash tests/parsers/check_additivity.sh develop HEAD`. It snapshots
both refs and prints a per-fixture `+added -removed ~changed` summary; the script
restores your checkout's editable install on exit. Latest local run of `develop`
vs this branch: all 15 fixtures identical, `PASS`.

## Status

- [x] Ready for review

## Breaking Changes

- [x] None

## Dependencies / Blockers

- [x] External dependencies (explain): `nomad-lab` pre-release from the MPCDF registry (snapshot-ordering determinism, nomad-FAIR MR 3103)

## Testing / Validation

- [x] Manual testing (explain): `develop` vs branch reports 15/15 identical, zero removed leaves, `PASS`; run deterministically via the MPCDF `nomad-lab` pre-release


ndaelman-hu:
**Heads-up: this workflow can't be dispatched from Actions until it's merged to `develop`.**

`snapshot-tests.yml` uses `workflow_dispatch`, and GitHub only registers manually-dispatchable workflows from the repo's **default branch** (`develop`). While the file exists only on this branch, `gh workflow run` and the "Run workflow" button return `404 … not found on the default branch`. Once this merges, the file is on `develop` and becomes dispatchable against any ref (the dispatched ref's version of the YAML is what runs).

In the meantime it's validated locally — the workflow is a thin wrapper around a script anyone can run:

```
bash tests/parsers/check_additivity.sh develop HEAD    # target, source
```

which runs the identical flow (snapshot both refs via worktree + `--no-deps`, then `compare_snapshots.py`). Latest local run vs `develop`: all 15 fixtures identical, `PASS`.

If it should run on the PR itself before merge, I can add a `pull_request` trigger — those run from the PR head and don't need the default branch.

tests/parsers/test_migration_snapshots.py
```diff
@@ -0,0 +1,197 @@
...
+)
+from nomad_simulation_parsers.parsers.vasp.parser import VASPParser
+from nomad_simulation_parsers.parsers.wannier90.parser import Wannier90Parser
+
+
+def serialize_archive_data(archive):
+    """Convert archive.data to dict for snapshot comparison."""
+    if not archive.data:
+        return None
+    return archive.data.m_to_dict(with_meta=False)
```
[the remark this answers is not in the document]
ndaelman-hu: The check compares two refs, so identical `None` on both is not a regression. A parser that genuinely fails to populate `archive.data` is caught by its own `test_<parser>_parser.py`; this harness only guards that the source does not *lose* data present in the target, where `None -> None` is correctly a non-event.

tests/parsers/test_migration_snapshots.py
```diff
@@ -0,0 +1,197 @@
...
+
+# ============= VASP =============
+
+
+def test_vasp_vasprun_snapshot(snapshot):
+    """Snapshot of VASP vasprun.xml parsing output."""
+    parser = VASPParser()
+    archive = EntryArchive()
+    parser.parse('tests/data/vasp/AgAc_relax/vasprun.xml.relax', archive, None)
+    assert serialize_archive_data(archive) == snapshot
```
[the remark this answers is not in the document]
ndaelman-hu: Fixed in 238c02b: a module-level `LOGGER = get_logger(__name__)` is now passed to every `parser.parse()` call instead of `None`.

tests/parsers/compare_snapshots.py
```diff
@@ -0,0 +1,122 @@
...
+            buffer.append(line)
+    if current is not None:
+        sections[current] = buffer
+
+    namespace = {'dict': dict, 'list': list, '__builtins__': {}}
+    result = {}
+    for name, lines in sections.items():
+        body = '\n'.join(lines).strip()
+        result[name] = eval(body, namespace) if body else None  # noqa: S307
+    return result
```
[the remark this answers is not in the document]
ndaelman-hu: The `.ambr` files are generated by syrupy in the same job, not untrusted input. `eval` runs with `__builtins__` stripped and only `dict`/`list` bound, over bodies that are exactly `dict(...)`/`list(...)`; the `# noqa: S307` documents the deliberate choice. An AST allowlist parser is possible but adds complexity for input we produce ourselves, so keeping as is.

tests/parsers/compare_snapshots.py
```diff
@@ -0,0 +1,122 @@
...
+    try:
+        return abs(float(a) - float(b)) <= 1e-9 * max(1.0, abs(float(a)))
+    except (TypeError, ValueError):
+        return False
+
+
+def main() -> int:
+    baseline = parse_ambr(sys.argv[1])
+    current = parse_ambr(sys.argv[2])
+
```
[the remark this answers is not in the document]
ndaelman-hu: Fixed in 238c02b: `main()` now prints a usage line and returns 2 when the two snapshot paths are missing, instead of raising `IndexError`.

.github/workflows/snapshot-tests.yml
```diff
@@ -0,0 +1,81 @@
...
+    - name: Set up Python 3.11
+      uses: actions/setup-python@v5
+      with:
+        python-version: 3.11
+    - name: Install uv
+      run: curl -LsSf https://astral.sh/uv/install.sh | sh
+    - name: Install dependencies
+      run: |
+        pip install --upgrade pip
+        uv pip install -e '.[dev,md]'
```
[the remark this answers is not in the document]
ndaelman-hu: This matches how the repo already installs `uv` (`actions.yml` uses the same `curl ... | sh`). Pinning the installer is a repo-wide policy choice better applied consistently across all workflows than in this one job.

tests/parsers/test_migration_snapshots.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+# ============= Quantum Espresso =============
+
+
+def test_quantumespresso_snapshot(snapshot):
+    """Snapshot of Quantum Espresso parsing output."""
+    parser = QuantumEspressoParser()
+    archive = EntryArchive()
+    parser.parse('tests/data/quantumespresso/pwscf/TiO2_opt/pw.out', archive, None)
```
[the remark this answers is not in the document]
ndaelman-hu: The committed version passes `ruff check` (E501 clean); this flagged a transient state.

tests/parsers/compare_snapshots.py
```diff
@@ -0,0 +1,197 @@
...
+        elif current is not None:
+            body.append(line)
+    if current is not None:
+        sections[current] = body
+
+    namespace = {'dict': dict, 'list': list, '__builtins__': {}}
+    return {
+        name: (eval('\n'.join(lines).strip(), namespace) if lines else None)  # noqa: S307
+        for name, lines in sections.items()
+    }
```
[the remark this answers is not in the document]
ndaelman-hu: `ruff check` passes on the committed version — no E501 on this line.

tests/parsers/compare_snapshots.py
```diff
@@ -0,0 +1,197 @@
...
+    if out is None:
+        out = {}
+    if isinstance(obj, dict):
+        for key, value in obj.items():
+            flatten(value, f'{path}.{key}', out)
+    elif isinstance(obj, list):
+        for index, value in enumerate(obj):
+            flatten(value, f'{path}[{index}]', out)
+    else:
+        out[path] = obj
```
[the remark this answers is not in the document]
ndaelman-hu: Fixed in 1e91992: `flatten()` now treats a wholly-empty archive (`archive.data is None`) as no leaves, so a None-vs-populated pair is additive rather than a false removal.

tests/parsers/compare_snapshots.py
```diff
@@ -0,0 +1,197 @@
...
+    print(
+        f'  {diff.verdict:10} {name:34} '
+        f'+{len(diff.added)} -{len(diff.removed)} ~{len(diff.changed)}'
+    )
+    for path in diff.removed[:_MAX_REMOVED_SHOWN]:
+        print(f'      removed: {path} = {base[path]!r}')
+    for path in diff.changed[:_MAX_CHANGED_SHOWN]:
+        print(f'      changed: {path}: {base[path]!r} -> {cur[path]!r}')
+    if len(diff.changed) > _MAX_CHANGED_SHOWN:
+        print(f'      changed: ... and {len(diff.changed) - _MAX_CHANGED_SHOWN} more')
```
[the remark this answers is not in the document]
ndaelman-hu: `ruff check` passes on the committed version — no E501 here.

tests/parsers/compare_snapshots.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+def main() -> int:
+    baseline = parse_ambr(sys.argv[1])
+    current = parse_ambr(sys.argv[2])
+
```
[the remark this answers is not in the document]
ndaelman-hu: Fixed in 238c02b: `main()` prints a usage line and returns 2 when the two paths are missing.

tests/parsers/compare_snapshots.py
```diff
@@ -0,0 +1,197 @@
...
+# 1. Parsing
+# --------------------------------------------------------------------------- #
+def parse_ambr(path: str) -> dict[str, Any]:
+    """Parse a syrupy amber file into ``{test_name: value}``.
+
+    The amber format groups each snapshot under a ``# name: <test>`` header,
+    followed by an indented body such as ``dict({...})`` / ``list([...])``.
+    Those bodies are valid Python once ``dict`` and ``list`` are in scope, so
+    they are evaluated in a namespace restricted to those two builtins. The
+    files are self-generated CI artifacts, so this is trusted input.
```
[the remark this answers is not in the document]
ndaelman-hu: The `.ambr` files are generated by syrupy in the same job, not untrusted input; `eval` runs with `__builtins__` stripped and only `dict`/`list` bound over bodies that are exactly `dict(...)`/`list(...)`, documented with `# noqa: S307`. Keeping as is.

tests/parsers/check_additivity.sh
```diff
@@ -0,0 +1,69 @@
...
+  ( cd "$wt" && python -m pytest --snapshot-update -q -p no:cacheprovider \
+      tests/parsers/test_migration_snapshots.py )
+  cp "$wt/tests/parsers/__snapshots__/test_migration_snapshots.ambr" "$out"
+  git -C "$REPO" worktree remove --force "$wt"
+}
+
+TARGET_SHA="$(resolve "$TARGET_REF")"
+SOURCE_SHA="$(resolve "$SOURCE_REF")"
+echo "source ${SOURCE_REF} (${SOURCE_SHA})  ->  target ${TARGET_REF} (${TARGET_SHA})"
+
```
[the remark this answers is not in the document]
ndaelman-hu: Fixed in 1e91992: the status line now reads "comparing source ... against target ..." to match the comparison direction.

.github/workflows/snapshot-tests.yml
```diff
@@ -0,0 +1,89 @@
...
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0  # full history so arbitrary source/target refs resolve
+
+    - name: Set up Python 3.11
+      uses: actions/setup-python@v5
+      with:
+        python-version: 3.11
+    - name: Install uv
+      run: curl -LsSf https://astral.sh/uv/install.sh | sh
```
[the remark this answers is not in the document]
ndaelman-hu: Matches how the repo already installs `uv` (`actions.yml` uses the same `curl ... | sh`); pinning the installer is a repo-wide policy choice better applied consistently across all workflows.

tests/parsers/test_migration_snapshots.py
```diff
@@ -0,0 +1,197 @@
+"""Snapshot-based verification for PR #170 migration.
+
+Run at baseline (1c7acd6) and current (test-data-normalization) to verify
+migration only adds properties, doesn't modify existing ones.
+"""
```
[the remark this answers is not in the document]
ndaelman-hu: By design: this module is excluded from the default suite (`actions.yml` runs pytest with `--ignore=tests/parsers/test_migration_snapshots.py`) and is invoked only via `check_additivity.sh`, which always passes `--snapshot-update` to regenerate. Snapshots are intentionally untracked and compared across two refs, so a bare `pytest` over this file is not a supported entrypoint.

ndaelman-hu:
@ladinesa as a heads-up, this test would be very helpful in #221. So, unless you have any major objections, I will add this as a manual CI/CD action. Screenshots are generated for the source and target branches, but never committed.

I will resolve the merge conflicts, then ping you again for a review.

ladinesa:
I still do not understand the use of syrupy here? Can't we simply use a json hash or direct comparison?

ndaelman-hu:
Good point. I originally reached for `syrupy` because it's a fast snapshot solution — quick at updating the `.ambr` files. But that only pays off if we retain and update snapshots, and here we generate them from scratch each run and don't keep them, so I agree the plain dump is the better fit.

I wouldn't dump a hash, though — we'll most likely want to see *where* the changes were applied, and a hash only tells us same-vs-different. A JSON dump (`archive.data.m_to_dict()`, already serializable) compared leaf-by-leaf gives us that.

And the diff stays additive: a purely additive change shouldn't be flagged — only leaves that actually modify between the two snapshots.

So I'm fine dropping `syrupy` and taking this lower down the stack.

ladinesa:
Here are my other suggestions:

1. Clarify and document the comparison policy. Maybe also strengthen it? fail on removed path and report changed value?
2. How to deal with intentional path moves and value changes?
3. Maybe rename from `test_migration_snapshots.py` and its workflow yml file to something less generic and descriptive e.g. compatibility_check? additive -> no_paths_removed?
4. I recommend to reuse exisiting fixtures in the integration tests.
5. Add another label e.g. compatiblity? so we can add it to nightly schedule
6. Include the harness hash, serializer version, python version uv lock hash, fixture hashes to the snapshot keys.

ladinesa:
> Good point. I originally reached for `syrupy` because it's a fast snapshot solution — quick at updating the `.ambr` files. But that only pays off if we retain and update snapshots, and here we generate them from scratch each run and don't keep them, so I agree the plain dump is the better fit.
> 
> I wouldn't dump a hash, though — we'll most likely want to see _where_ the changes were applied, and a hash only tells us same-vs-different. A JSON dump (`archive.data.m_to_dict()`, already serializable) compared leaf-by-leaf gives us that.
> 
> And the diff stays additive: a purely additive change shouldn't be flagged — only leaves that actually modify between the two snapshots.
> 
> So I'm fine dropping `syrupy` and taking this lower down the stack.

I meant first a fast hash check then do the comparison.

tests/parsers/generate_snapshots.py
```diff
@@ -0,0 +1,114 @@
...
+    'crystal': [(CrystalParser, 'tests/data/crystal/single_point/dft/output.out')],
+    'fhiaims': [(FHIAimsParser, 'tests/data/fhiaims/Si_geomopt/out.out')],
+    'gromacs': [(GromacsParser, 'tests/data/gromacs/fe_test/md.log')],
+    'h5md': [
+        (
+            H5MDParser,
+            'tests/data/h5md/test_traj_openmm_reduced-SOL_5frames_07-10-25.h5',
+        )
+    ],
+    'lammps': [(LammpsParser, 'tests/data/lammps/1_xyz_files/log.lammps')],
```
ladinesa: lobster and yambo are not happy to be left out

ndaelman-hu:
Thanks — going point by point.

On the hash: agreed, a fast full-snapshot hash as a short-circuit, then the leaf-by-leaf diff only when they differ. I'll add that.

1. Agreed, I'll document it — the policy lives in the script header, the workflow description, and the docs. In short, three tiers: `identical` (passes silently), `additive` (passes, but the added paths are surfaced as a non-failing warning annotation), and removed-or-changed (hard fail, printing `old -> new`). Additions stay visible without turning the check red; any removal or value change blocks.

2. A deliberate path move or value change lands in the hard-fail tier from point 1 (removed / changed) — by design, not something to suppress. Since the gate is manual-only and never wired into the PR CI, you run the script locally, read the same `old -> new` diff, and confirm the change is what you intended. That local run is the review step, so I'd rather not add a waiver mechanism to the tool.

3. Agreed on dropping "migration". I like `additive` better than `no_paths_removed`, so I kept it and just appended qualifiers for clarity — the workflow becomes `additive-output-changes` (`compatibility` reads too vague to me). It also fits the three tiers: `additive` names the middle signal specifically, whereas `no_paths_removed` would cover both the `identical` and `additive` tiers.

4. Agreed — and you're right that lobster and yambo were left out. Lobster I'll add now; it already has integration fixtures (e.g. `tests/data/lobster/NaCl/lobsterout`). Yambo is the real gap: it has no end-to-end fixture in the checkout — the tests are unit-only and there's no `tests/data/yambo/` — so there's nothing to snapshot until a small mainfile is added. I'll wire the generator to the integration fixtures (which is exactly what surfaces this), and open a separate issue covering both the shared fixture registry and giving yambo an end-to-end fixture.

5. I'd keep this manual-only and off the nightly schedule — the baseline is a source/target ref pair chosen per run, which doesn't map cleanly onto a scheduled job.

6. Agreed — I'll fold the Python version, `uv.lock` hash, and fixture hashes into the snapshot cache key / provenance header, so a stale environment can't produce a false match.

I'll also fix the two Copilot nits: quoting the `parsers` input as a single argument, and using the `mktemp` directory directly so no empty temp dir is left behind.

tests/parsers/generate_snapshots.py
ndaelman-hu: Lobster added; yambo has no end-to-end fixture yet (unit-only, no `tests/data/yambo/`), so I am tracking it in a separate issue.

tests/parsers/compare_snapshots.py
```diff
@@ -0,0 +1,234 @@
...
+    ``.outputs[0].total_energies[0].value``. Only scalar leaves are stored, so
+    two archives can be compared by set operations on their leaf paths.
+    """
+    if out is None:
+        out = {}
+    if isinstance(obj, dict):
+        for key, value in obj.items():
+            flatten(value, f'{path}.{key}', out)
+    elif isinstance(obj, list):
+        for index, value in enumerate(obj):
```
ladinesa: is there a way we can support reordering?

tests/parsers/compare_snapshots.py
```diff
@@ -0,0 +1,234 @@
...
+        document = json.load(handle)
+    if isinstance(document, dict) and 'snapshots' in document:
+        return document['snapshots'], document.get('provenance', {})
+    return document, {}
+
+
+# --------------------------------------------------------------------------- #
+# 2. Flattening
+# --------------------------------------------------------------------------- #
+def flatten(
```
ladinesa: I suggest you also add record type and length

ndaelman-hu:
Tried this locally against #221 (the ORCA transformer refactor) as a real-world check: `check_additivity.sh develop orca-transformer-refactor orca` reports the source snapshots identical to the target (hash short-circuit), matching the independent byte-identical result for that refactor. So the harness correctly clears a behavior-preserving refactor.

tests/parsers/check_additivity.sh
```diff
@@ -0,0 +1,96 @@
...
+  fi
+  local tmp wt
+  tmp="$(mktemp -d)"
+  wt="$tmp/snapshot-${name}"       # git worktree add needs a non-existent path
+  git -C "$REPO" worktree add --quiet --force "$wt" "$sha"
+  cp "$HERE/generate_snapshots.py" "$wt/tests/parsers/"
+  uv pip install -e "$wt" --no-deps >/dev/null
+  # shellcheck disable=SC2086  # word-splitting of $PARSERS is intended
+  ( cd "$wt" && python tests/parsers/generate_snapshots.py "$out" $PARSERS )
+  git -C "$REPO" worktree remove --force "$wt"
```
ladinesa: this is not called if failure happens before this. I suggest you use a temporary env

tests/parsers/compare_snapshots.py
ndaelman-hu: Possible, but it needs a structural multiset check per list rather than the index diff, surfaced as a distinct `reordered` tier -- never silent, since order here is often meaningful (SCF/opt steps, frames). Given the nomad-lab pre-releases already give deterministic ordering, are you actually hitting legitimate reordering? If so I will add it; otherwise the fixed ordering covers it.

ladinesa: ok, we implement it when the need arises.

tests/parsers/compare_snapshots.py
ndaelman-hu: Done -- list length is now an explicit `[len]` leaf, so a resize or an empty-vs-absent list is caught directly rather than only via index removal. Leaf comparison is also type-sensitive now: int/float noise is tolerated, but a value flipping type (e.g. `1 -> "1"`, `0 -> null`) is flagged rather than slipping through `==`.

tests/parsers/check_additivity.sh
ndaelman-hu: Done. Since the generator imports the parser classes directly rather than through entry points, I dropped the install-and-restore entirely: each ref now runs with its worktree `src` on `PYTHONPATH`, so the caller environment is never touched and there is nothing to restore on failure. Verified the ORCA refactor still compares byte-identical, and that the schema code (not just the top package) resolves from the worktree.

tests/parsers/compare_snapshots.py
ladinesa: this is still identical {'data': {}}  versus  {} i suggest you add .data[type] = 'dict'

ndaelman-hu: Added -- every container now emits a `[type]` leaf, so `{'data': {}}` no longer matches `{}` (it flags `.data[type]` removed).

ndaelman-hu: For completeness: `dict` and `list` are the only JSON containers, so those are the only `[type]` markers. The remaining JSON types -- `str`, `int`, `float`, `bool`, `null` -- are scalar leaves already covered by the type-sensitive comparison (`0 -> null`, `1 -> "1"`, etc. all flag).

.github/workflows/additive-output-changes.yml
```diff
@@ -0,0 +1,101 @@
...
+          echo "target_sha=$(resolve "$TARGET")"
+          echo "source_sha=$(resolve "$SOURCE")"
+        } >> "$GITHUB_OUTPUT"
+
+    # Cache each ref's snapshot on its SHA + the requested parser set + the
+    # environment (python version, uv.lock hash); check_additivity.sh reuses any
+    # .json that already exists, so a cache hit skips that ref's regeneration.
+    # A changed environment misses the cache and regenerates, matching the
+    # provenance recorded inside the snapshot.
+    - name: Restore cached target snapshot
```
ladinesa: Could this lead to a situtation where you have the same target_sha and uv.lock and return a target_snapshots.json for a different source which you then reuse in the script?

.github/workflows/additive-output-changes.yml
```diff
@@ -0,0 +1,101 @@
...
+      uses: actions/cache@v4
+      with:
+        path: target_snapshots.json
+        key: snapshot-${{ steps.refs.outputs.target_sha }}-${{ inputs.parsers }}-py311-${{ hashFiles('uv.lock') }}-v3
```
ladinesa: should you also add the hash for the generator file

ladinesa:
Address the last two points then merge.

ladinesa (approved):

.github/workflows/additive-output-changes.yml
ndaelman-hu: Good catch -- yes, with autodetect the parser set comes from the target..source diff, so that reuse was possible. Fixed: both cache keys now include the target and source SHAs, so a snapshot is only reused for the same ref pair.

.github/workflows/additive-output-changes.yml
ndaelman-hu: Done -- the generator script is now folded into the cache-key hash (combined with uv.lock), so a changed generator invalidates stale snapshots.

ndaelman-hu:
> Address the last two points then merge.

Thank you very much for all the feedback! I like how you went into the design cosndirations and various failure modes I hadn't considered yet. To codify this insight, I also opened an issue to add some tests for this framework (#225).

Just finished implementing your last points and polished a bit more myself. Merging now.

PR #198, opened 2026-07-30
Gromacs parser fix model system blowing up in size

Bernadette-Mohr:
I had mistakenly appended non-changing quantities (lattice_vectors, periodic_boundaries, bond_list, particle_states, ...), to every trajectory frame, blowing up the archive size. This is fixed here, immutable quantities are only appended once to the top-level model system.

The error flagged by ruff is not in the gromacs parser.


Bernadette-Mohr:
branch based off of wrong state, redoing cleanly from develop.

PR #196, opened 2026-07-27
Parse the VASP DFT exchange-correlation functional and Jacob's ladder

ndaelman-hu:
## Purpose

The VASP parser did not populate the DFT method's exchange-correlation functional: the `DFT.xc` subsection was never wired (its mapping was commented out), so `archive.data.model_method[].xc` came out empty and `jacobs_ladder` was `unavailable`. This PR adds XC parsing end-to-end. Rather than hand-maintain a per-code tag-to-LibXC-label table, the parser emits a single canonical `functional_key` derived from the VASP `GGA`/`METAGGA`/hybrid tags, and the `nomad-simulations` schema expands it into complete, registry-validated `components` (with `family`/`kind`) and derives `jacobs_ladder`. Completeness lives once in the schema's shared alias/registry tables.

## Scope

**Included**

- Bind the `DFT.xc` subsection for both the vasprun.xml and OUTCAR sources (previously unmapped) so XC is parsed at all.
- Populate it via `get_functional_key` (`VASP_TAG_TO_FUNCTIONAL`, GGA/METAGGA/hybrid to canonical name) mapped onto `XCFunctional.functional_key`; the schema does the LibXC expansion and `jacobs_ladder` derivation.
- vasprun.xml: recursively walk `<parameters>`, typing values by the `<i type=...>` attribute — required to reach `GGA`/`METAGGA`/`LHFCALC` across nested separators.
- Centralize unset-tag normalization in `_clean_tag` (OUTCAR bool `False`, absent elements, `--`/`NONE` sentinels).
- Parametrized test over both sources.

**Out of scope**

- Migrating the other simulation parsers (QuantumESPRESSO, exciting, ...) to `functional_key` — they still emit components.
- Hybrid exact-exchange mixing into `XCFunctional.global_exact_exchange` (left as a `TODO`; needs a hybrid fixture).

## Context & Links

- Depends on the companion aliases PR: FAIRmat-NFDI/nomad-simulations#435 (must land first/together).

## Reviewer Notes

- `XCFunctional.normalize` only expands `functional_key` when `components` is empty, so the parser must not also emit components.
- Additive and verified non-destructive on a real vasprun (`case_1`): the `data` section is byte-identical to a pre-change run modulo timestamps/ordering, except `xc`/`jacobs_ladder` are now populated (`functional_key=PBE`, `jacobs_ladder=GGA`, `XC_GGA_X_PBE` + `XC_GGA_C_PBE`).
[...]

PR #199, opened 2026-07-30
Extend DFT exchange-correlation / Jacob's ladder parsing to the remaining parsers

ndaelman-hu:
## Purpose

Extends the VASP exchange–correlation and Jacob's ladder work (#196) to the
other DFT parsers, so each populates `DFT.xc` and lets the schema derive
`jacobs_ladder`. Parsers hand the schema a standard functional name where the
code names the functional, or the raw LibXC components where the code reports
the individual exchange/correlation parts — never doing LibXC taxonomy lookups
themselves.

## Scope

**Included** — XC parsing for eight parsers, one commit each:

| Parser | Representation | add / rework |
|---|---|---|
| fhiaims | `functional_key` (name from XC control string) | add |
| gpaw | `functional_key` (native name) | add |
| ams | `functional_key` (name of the highest active rung) | add |
| quantumespresso | `functional_key` (printed name or four-slot code; XML `<functional>`) | add |
| octopus | free-form components (LibXC labels) | add |
| abinit | free-form components (LibXC labels / ids from `ixc`) | rework |
| crystal | free-form components (LibXC labels) | rework |
| exciting | free-form components (LibXC label / INFO id) | rework |

Name-path parsers set `XCFunctional.functional_key`; the schema expands it via
the existing alias table. Component-path parsers populate `XCFunctional.components`
with raw LibXC labels/ids; the schema resolves the taxonomy.

**Out of scope**

- VASP itself (this builds on #196).
- Non-DFT parsers (gromacs, lammps, h5md, lobster, phonopy, wannier90, yambo).

## Context & Links

- Targets #196 — same `functional_key` / component pattern, extended.
- Depends on nomad-simulations PR #440 (free-form component resolution + the
  `unidentified` marker) for the four component-path parsers. That PR should
  merge first.

## Reviewer Notes

- Representation per parser reflects what the code actually reports: a name when
  the code names the functional, raw LibXC parts (label or id) when it does not.
  This keeps LibXC knowledge in the schema, not the parsers.
- Fixes surfaced while wiring these up:
  - octopus dropped the exchange component because its table keyed only the
    legacy `Exchange` spelling, not the `Slater exchange` the code prints.
  - crystal emitted no method at all: the `components` transformer sourced
    `.dft` while `DFT` is rooted at `.dft`, resolving to `.dft.dft` (nothing),
    so the empty `DFT` was pruned. Now sourced from `.@`.
  - exciting always re-read `input.xml` because its guard tested the removed
    `xc_functionals` field; repointed to `xc.components`.
  - abinit keeps every reported component: a `?` slot becomes an `unidentified`
    placeholder and an out-of-table LibXC id is handed to the schema as a raw
    id, rather than dropping it.

## Status

- [x] Ready for review (pending #440)

## Breaking Changes

- [x] None.

## Dependencies / Blockers

- [x] nomad-simulations PR #440 — required by the four component-path parsers.

## Testing / Validation

- [x] Unit tests — each parser's suite passes.
- [x] Manual testing — sample-parsed a fixture per parser; confirmed
  `functional_key`/`components` resolve to `family`/`kind` and the expected
  `jacobs_ladder` rung.

### DIFF
--- a/src/nomad_simulation_parsers/parsers/fhiaims/parser.py
+++ b/src/nomad_simulation_parsers/parsers/fhiaims/parser.py
@@ -68,112 +68,39 @@ class FHIAimsOutMappingParser(TextMappingParser):
         'scgw': 'scGW',
     }
 
-    _xc_map = {
-        'Perdew-Wang parametrisation of Ceperley-Alder LDA': [
-            {'name': 'LDA_C_PW'},
-            {'name': 'LDA_X'},
-        ],
-        'Perdew-Zunger parametrisation of Ceperley-Alder LDA': [
-            {'name': 'LDA_C_PZ'},
-            {'name': 'LDA_X'},
-        ],
-        'VWN-LDA parametrisation of VWN5 form': [
-            {'name': 'LDA_C_VWN'},
-            {'name': 'LDA_X'},
-        ],
-        'VWN-LDA parametrisation of VWN-RPA form': [
-            {'name': 'LDA_C_VWN_RPA'},
-            {'name': 'LDA_X'},
-        ],
-        'AM05 gradient-corrected functionals': [
-            {'name': 'GGA_C_AM05'},
-            {'name': 'GGA_X_AM05'},
-        ],
-        'BLYP functional': [{'name': 'GGA_C_LYP'}, {'name': 'GGA_X_B88'}],
-        'PBE gradient-corrected functionals': [
-            {'name': 'GGA_C_PBE'},
-            {'name': 'GGA_X_PBE'},
-        ],
-        'PBEint gradient-corrected functional': [
-            {'name': 'GGA_C_PBEINT'},
-            {'name': 'GGA_X_PBEINT'},
-        ],
-        'PBEsol gradient-corrected functionals': [
-            {'name': 'GGA_C_PBE_SOL'},
-            {'name': 'GGA_X_PBE_SOL'},
-        ],
-        'RPBE gradient-corrected functionals': [
-            {'name': 'GGA_C_PBE'},
-            {'name': 'GGA_X_RPBE'},
-        ],
-        'revPBE gradient-corrected functionals': [
-            {'name': 'GGA_C_PBE'},
-            {'name': 'GGA_X_PBE_R'},
-        ],
-        'PW91 gradient-corrected functionals': [
-            {'name': 'GGA_C_PW91'},
-            {'name': 'GGA_X_PW91'},
-        ],
-        'M06-L gradient-corrected functionals': [
-            {'name': 'MGGA_C_M06_L'},
-            {'name': 'MGGA_X_M06_L'},
-        ],
-        'M11-L gradient-corrected functionals': [
-            {'name': 'MGGA_C_M11_L'},
-            {'name': 'MGGA_X_M11_L'},
-        ],
-        'TPSS gradient-corrected functionals': [
-            {'name': 'MGGA_C_TPSS'},
-            {'name': 'MGGA_X_TPSS'},
-        ],
-        'TPSSloc gradient-corrected functionals': [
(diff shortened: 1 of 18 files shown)


src/nomad_simulation_parsers/parsers/fhiaims/parser.py
```diff
@@ -68,112 +68,39 @@ def logger(self):
...
+    # FHI-aims XC control description -> standard functional name. The
+    # `nomad-simulations` schema expands the name into LibXC components
+    # (family/kind) and derives `jacobs_ladder`; the parser deliberately does
+    # not resolve LibXC labels itself. Names normalize case-insensitively, so
+    # the human-readable spelling used here resolves to the schema alias.
+    _xc_name_map = {
+        'Perdew-Wang parametrisation of Ceperley-Alder LDA': 'LDA',
+        'Perdew-Zunger parametrisation of Ceperley-Alder LDA': 'PZ81',
+        'VWN-LDA parametrisation of VWN5 form': 'VWN',
+        'VWN-LDA parametrisation of VWN-RPA form': 'VWN',
```
EBB2675: why do non-rpa and rpa versions map to the same value? I think we should distinguish the latter as VWN-RPA

and a coordinated addition to the schema in PR 440:
`"VWN-RPA": ["XC_LDA_X", "XC_LDA_C_VWN_RPA"],`

EBB2675:
```suggestion
        'VWN-LDA parametrisation of VWN-RPA form': 'VWN-RPA',
```

EBB2675 (commented):
We previously preserved the RPA variant as `LDA_C_VWN_RPA`, but mapping it to VWN now resolves to the non-RPA `XC_LDA_C_VWN`. Could we retain a distinct key for this variant and add the corresponding alias in `nomad-simulations`?

src/nomad_simulation_parsers/parsers/fhiaims/parser.py
EBB2675: added to schema PR 440 as a suggestion block

ndaelman-hu: Done — mapped the VWN-RPA control string to `VWN-RPA` in 490026c, and added the coordinated `"VWN-RPA": ["XC_LDA_X", "XC_LDA_C_VWN_RPA"]` alias to PR #440 (f314031). `VWN` (VWN5) is left unchanged; `VWN-RPA` now resolves to the RPA correlation `XC_LDA_C_VWN_RPA`.

src/nomad_simulation_parsers/parsers/quantumespresso/parser.py
```diff
@@ -144,80 +112,39 @@ def get_energy_contributions(self, source: dict[str, Any]):
...
+        tokens = source.split('(', 1)[0].split()
+        if not tokens:
+            return None
+        if len(tokens) == 1:
+            return tokens[0]
+        combo = ' '.join(tokens)
+        name = self._slot_combo_names.get(combo)
+        if name is None:
+            self.logger.debug('unmapped QE XC slot combination', data=dict(value=combo))
+        return name
```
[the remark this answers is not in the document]
ndaelman-hu: Done in e814470 — falls back to the raw slot combination instead of returning `None`, so the reported XC is preserved (consistent with the single-token path).

src/nomad_simulation_parsers/parsers/fhiaims/parser.py
ndaelman-hu: Applied

ndaelman-hu:
@EBB2675 I will merge this after you have reviewed its target too: else it will be hard to see the individual changes.

EBB2675 (approved):
lgtm 👍

PR #207, opened 2026-08-03
Remove special handling of electronic dos and bs

ladinesa:


src/nomad_simulation_parsers/parsers/quantumespresso/pwscf/parser.py
```diff
@@ -439,156 +439,46 @@ def build_workflow(self):
...
-                if output.electronic_dos and reference_energy is not None:
-                    for dos in output.electronic_dos:
-                        if dos.energies_origin is None:
-                            dos.energies_origin = reference_energy
-
-        # TODO(mapping-migration): remove this parser-side forces fallback once
-        # per-step force payloads can be expressed via mapping annotations.
-        # `Outputs.total_forces` maps '.@', which lost the step-resolved forces
-        # when `get_configurations` started returning only the final SCF step.
-        if hasattr(self.mainfile_parser, 'get_configuration_forces'):
```
ndaelman-hu, line 533: This fallback is what kept the per-step forces. On `TiO2_opt/pw.out`, `develop` gives 4 `total_forces` on `outputs[1]` (one per ionic step); this branch gives 1, only the final step.

This happens because `Outputs.total_forces` maps `.@` and `get_configurations` only returns the final SCF step. That is the data loss #170 fixed. `get_configuration_forces` is now unused. Suggest mapping it, or keeping the fallback for now.

I suggest adding a test that counts the no. `total_forces` to prevent future regressions.

src/nomad_simulation_parsers/schema_packages/quantumespresso/pwscf.py
```diff
@@ -84,13 +90,10 @@ class Outputs(outputs.Outputs):
...
-        '.electronic_band_structures',
-    )
-    add_mapping_annotation(
-        outputs.Outputs.electronic_dos,
-        OUT_KEY,
-        '.electronic_dos',
+        ('get_eigenvalues', ['.@']),
     )
+    add_mapping_annotation(outputs.Outputs.electronic_dos, DOS_KEY, '.@')
+    add_mapping_annotation(outputs.ElectronicDensityOfStates.m_def, OUT_KEY, '.@')
```
ndaelman-hu: This annotation can trigger a _sparse DOS_ with just `energies_origin` but NO `value`, or `energies`. You can check `TiO2_opt` and `H2O+_scf.out`.
Such nonsensical DOS should never be instantiated. You can try adapting the annotations or extending `get_reference_energy` to require all 3 parameters.

Pls add another count test, to avoid future regressions.

src/nomad_simulation_parsers/schema_packages/quantumespresso/pwscf.py
```diff
@@ -24,17 +24,22 @@ class TotalForce(outputs.TotalForce):
...
 
 class ElectronicEigenvalues(outputs.ElectronicEigenvalues):
-    add_mapping_annotation(outputs.ElectronicEigenvalues.n_levels, OUT_KEY, '.n_levels')
     add_mapping_annotation(outputs.ElectronicEigenvalues.value, OUT_KEY, '.eigenvalues')
     add_mapping_annotation(
         outputs.ElectronicEigenvalues.occupation, OUT_KEY, '.occupations'
     )
+    add_mapping_annotation(outputs.ElectronicEigenvalues.n_levels, OUT_KEY, '.n_levels')
+    add_mapping_annotation(
+        outputs.ElectronicEigenvalues.spin_channel, OUT_KEY, '.spin_channel'
```
ndaelman-hu, line 33: `get_eigenvalues` does not return a `spin_channel` key, so this annotation (and the one on `ElectronicBandStructure`) does nothing.

On `H2O+_scf.out` both instances have distinct values but `spin_channel=None`. Either return `spin_channel` from the transformer or drop the annotation.

Careful with channel `0`: falsy values are dropped in `from_dict`.

src/nomad_simulation_parsers/parsers/quantumespresso/pwscf/parser.py
```diff
@@ -439,156 +439,46 @@ def build_workflow(self):
...
+        return LOGGER
+
+    def load_file(self):
+        for dos_file in search_files(
+            pattern='*.dos', basedir=os.path.dirname(self.filepath)
+        ):
+            self.text_parser.mainfile = dos_file
+            if self.text_parser.energies is not None:
+                break
+        return self.text_parser
```
ndaelman-hu, line 434: `_dos_parser` on `PWSCFArchiveWriter` is a class attribute, so every entry parsed in the same process reuses the same `DOSParser` and the same inner `text_parser`.

The highlighted loop only replaces that data when it finds a `*.dos` file. When `search_files` finds none, the loop body never runs, nothing is cleared, and `return self.text_parser` hands back the previous entry's DOS.

Repro: parse `W_dos/w.dos.out`, then `gipaw/scf_out_epr_out/H2O+_scf.out` in the same process. The H2O+ archive ends up with the 1801-point W DOS.

Fix: reset `text_parser` (clear `mainfile`/results) before the loop, and return early when no file matches.

ndaelman-hu (commented):
This PR  refactors the Quantum ESPRESSO PWSCF parser's imperative fallback code for eigenvalues, band structures, DOS, and forces, replacing it with declarative mapping annotations plus a new companion-file pass that reads `dos.x` output (`*.dos`) into `electronic_dos`.

I found four issues, see the inline comments. I would also extend the test coverage to automatically catch these regressions.

src/nomad_simulation_parsers/parsers/quantumespresso/pwscf/parser.py
ladinesa: In the earlier pr, you added only the last scf step to the configurations. I doubt if this is really the intention as all the geometry optimization steps now only yield one output section. For the `TiO2` case, should we not have 6 outputs sections in total, each one having 1 total_forces section not 2 outputs and the second output having 4 forces?

src/nomad_simulation_parsers/schema_packages/quantumespresso/pwscf.py
ladinesa: Add spin_channel, 0 seems to be set.

src/nomad_simulation_parsers/parsers/quantumespresso/pwscf/parser.py
ladinesa: set mainfile before loop

src/nomad_simulation_parsers/parsers/quantumespresso/pwscf/parser.py
ndaelman-hu: Agreed, your shape is the intended one. My earlier suggestion came from comparing against the parser's develop branch and treating its output as the baseline to preserve.
While the schema is suggestive of the 1:1-mapping, I recently tried to make it explicit in the `nomad-simulations` docs (`explanation/simulation_entry.md`). @JFRudzinski can probably confirm the intended mapping here.

So, to be clear, you are suggesting the following structure for `TiO2_opt`, correct? Cause this is not the one that is currently created.

```
model_system[0..5]      # initial, 4 bfgs steps, final scf
outputs[0..5]                 # index-aligned 1-1 with model_system (model_system_ref)
├── total_forces[0]   # one (n_atoms, 3) TotalForce for that configuration
└── scf_steps          # that configuration's own SCF convergence
```

src/nomad_simulation_parsers/schema_packages/quantumespresso/pwscf.py
ndaelman-hu: You're right, 0 comes through fine. I double-checked `from_dict`: the falsy-dropping I had in mind only kicks in when pruning empty sub-sections, individual quantities just go through m_set. So my warning doesn't apply here. All good.

ndaelman-hu: I did notice one issue: restricted runs now also get `spin_channel=0` (see `TiO2_opt`), which the schema currently interprets as spin-polarized. 

Given that even a developer is unaware of this rule, it'll become hard to enforce across parsers in the future. So perhaps I should read this as feedback to the schema rather than a parser problem. I'll take it to `nomad-simulations`. Leaving this as is then, though it's the opposite call from the DOSCAR in #208, so we should align the two once that's settled.

ndaelman-hu: issue opened in https://github.com/FAIRmat-NFDI/nomad-simulations/issues/459

src/nomad_simulation_parsers/schema_packages/quantumespresso/pwscf.py
ndaelman-hu: Thanks, the `DOS_OUT_KEY` split fixes this — I checked `TiO2_opt` and `H2O+`, no more empty DOS sections.

ndaelman-hu: @ladinesa One thing I noticed in the fix though: the new contributions -> ('get_dos_contributions', ['.@']) annotation doesn't do anything. The transformer lives on `DOSParser`, but `DOS_OUT_KEY` is converted by the mainfile parser, which doesn't have that method — so it never resolves. Moreover, it returns `[]` anyway.

src/nomad_simulation_parsers/schema_packages/quantumespresso/pwscf.py
ladinesa: will now write only if spin pol.

src/nomad_simulation_parsers/schema_packages/quantumespresso/pwscf.py
ladinesa: It was a fix because the nested reference to ElectronicDensityOfStates unintentionally creates contributions sections. I am actually not sure what contributions means for dos and there is already a sub-section for projected dos.

src/nomad_simulation_parsers/parsers/quantumespresso/pwscf/parser.py
ladinesa: corrected configurations parsing to reflect the structure above.

ndaelman-hu: Verified on TiO2_opt: 6 model_system / 6 outputs, 1-1 via model_system_ref, one total_forces each, per-step scf_steps. Exactly the structure above — thanks.

src/nomad_simulation_parsers/schema_packages/quantumespresso/pwscf.py
ndaelman-hu: Good point. I could drop `projected_dos` for `contributions`. Will open an issue to decide. You can leave as-is for this PR.

ndaelman-hu: https://github.com/FAIRmat-NFDI/nomad-simulations/issues/460

ndaelman-hu (approved):
All previous comments applied. You mentioned that you would add a guard for spin-pol, feel free to do so.

PR #208, opened 2026-08-05
Consolidate doscar parser

ladinesa:


JFRudzinski (approved):

src/nomad_simulation_parsers/parsers/vasp/doscar_parser.py
```diff
@@ -0,0 +1,171 @@
...
+        for atom_index, pdos_n in enumerate(pdos):
+            pdos_atom = np.transpose(pdos_n)
+            n_lm = np.shape(pdos_atom)[0] // n_spin
+            pdos_atom = np.reshape(pdos_atom, (n_lm, n_spin, n_dos))
+            pdos_dict = dict(
+                dos=np.transpose(pdos_atom, (1, 0, 2)), atom_index=atom_index
+            )
+            if n_lm == 1:
+                fields = ['s']
+            if n_lm == 3:
```
ndaelman-hu: The first `if` is not part of the chain below, so for `n_lm == 1` the `['s']` is overwritten by the `else` branch to `[None]`. Should be `elif n_lm == 3`. (Carried over from the old lobster parser; `dos_fields` has no consumer yet, but the value is wrong.)
```suggestion
            elif n_lm == 3:
```

src/nomad_simulation_parsers/parsers/vasp/doscar_parser.py
```diff
@@ -0,0 +1,171 @@
...
+        if not self.data_object:
+            return {}
+
+        self.data_object.parse()
+        return self.data_object._results or {}
+
+    def get_dos(
+        self, total: np.ndarray, projected: list[dict[str, Any]]
+    ) -> list[dict[str, Any]]:
+        n_spin = max(1, len(total) // 2)
```
ndaelman-hu: `total[n_spin + spin]` assumes the integrated-DOS columns are present. That is safe for spec-conforming files (the [DOSCAR format](https://www.vasp.at/wiki/index.php/DOSCAR) always includes them, and Lobster writes the same layout), but `DOSCARFileParser` explicitly supports minimal energy-and-DOS-only files (`n_spin = max(1, ...)` and the 'reduced test/minimal files' comment). Those files then fail here with an `IndexError` that the transformer swallows, so their DOS silently disappears. Either drop the minimal-file support in the file parser or handle the missing column here too.

src/nomad_simulation_parsers/schema_packages/vasp.py
```diff
@@ -461,7 +464,7 @@ class ElectronicDensityOfStates(outputs.ElectronicDensityOfStates):
         outputs.ElectronicDensityOfStates.value, XML2_KEY, '.value', unit='1/eV'
     )
     add_mapping_annotation(
-        outputs.ElectronicDensityOfStates.value, OUTCAR_KEY, '.value', unit='1/eV'
+        outputs.ElectronicDensityOfStates.value, DOSCAR_KEY, '.value', unit='1/eV'
```
ndaelman-hu:
```suggestion
```

This annotation is dead: line 500 annotates the same quantity under the same `DOSCAR_KEY` with `.dos`, and `add_mapping_annotation` is last-wins. Remove this one.

ndaelman-hu (commented):
This PR consolidates DOSCAR parsing into a shared `vasp/doscar_parser.py` used by both the VASP OUTCAR and Lobster parsers, replacing the imperative `get_total_dos` with a `DOSCAR_KEY` mapping pass.

Looks good overall: the four Lobster fixtures produce identical archives to `develop`, and `AgAc_relax/OUTCAR` keeps its total DOS while gaining projected DOS. Four smaller issues in the inline comments.

Notes:
- The fresh `DOSCARParser()` per parse with `load_file` returning `None` when no DOSCAR exists is the right state handling; the QE `DOSParser` in #207 would benefit from the same pattern.
- The old `get_total_dos` stored `np.abs(dos)` (the QE parser still does); here raw values pass through, so negative DOS artifacts now reach `ElectronicDensityOfStates.value`. Intentional?
- Follow-up idea: the vasprun path could gain projected DOS from the `<partial>` block, which carries the same data as the DOSCAR ion blocks plus explicit orbital labels, so no column-count inference is needed.
- Test suggestions: assert `dos.spin_channel is None` in `test_outcar_electronic_outputs_from_doscar_and_eigenvalues` (the fixture is spin-restricted), and add a 5-column variant to `test_outcar_doscar_suffix_resolution` to cover the spin-polarized branch of `get_dos` (spin 0/1 and the integrated-column split), which currently has no coverage.

src/nomad_simulation_parsers/schema_packages/vasp.py
```diff
@@ -59,6 +61,7 @@ class Simulation(general.Simulation):
     add_mapping_annotation(general.Simulation.outputs, OUTCAR_KEY, '.calculation')
     # TODO: make update_mode merge@last when mapping parser is updated
     add_mapping_annotation(VASPOutputs.m_def, CHGCAR_KEY, '.@', update_mode='append')
+    add_mapping_annotation(general.Simulation.outputs, DOSCAR_KEY, '.@')
```
ndaelman-hu: Without an `update_mode` the DOS merges into `outputs[0]`. For a relaxation OUTCAR (one output per ionic step) the DOSCAR belongs to the final step, so this wants `merge@last` like the CHGCAR TODO above and the QE DOS in #207. Fine for the current single-point fixture, but worth the same TODO at least.

src/nomad_simulation_parsers/parsers/vasp/doscar_parser.py
```diff
@@ -0,0 +1,171 @@
...
+                    value = np.array(line_split, dtype=np.float64)
+                except Exception:
+                    atomic_number = self.re_atomic_number.match(line)
+                    if atomic_number:
+                        atomic_numbers.append(int(atomic_number.group(1)))
+                    continue
+
+                if n == n_header:
+                    # VASP's DOSCAR header is Emax, Emin, NEDOS, Efermi.
+                    # Older Lobster files may omit the latter fields.
```
ndaelman-hu, line 52: Please back these statements with links, e.g.:
```suggestion
                    # VASP's DOSCAR header is Emax, Emin, NEDOS, Efermi
                    # (https://www.vasp.at/wiki/index.php/DOSCAR).
                    # Older Lobster files may omit the latter fields.
```
For the Lobster claim, add the user guide or the version where this changed; all `DOSCAR.lobster` test fixtures have the full five-field header, so the `value[1]` fallback is otherwise hard to verify.

src/nomad_simulation_parsers/parsers/vasp/doscar_parser.py
```diff
@@ -0,0 +1,171 @@
...
+    def get_dos(
+        self, total: np.ndarray, projected: list[dict[str, Any]]
+    ) -> list[dict[str, Any]]:
+        n_spin = max(1, len(total) // 2)
+        dos = []
+        for spin in range(n_spin):
+            dct = dict(
+                dos=total[spin],
+                integrated=total[n_spin + spin],
+                spin=spin,
```
ndaelman-hu: For spin-restricted files this sets `spin_channel=0` on the single summed-DOS section. The schema uses an unset `spin_channel` to mean not spin-polarized: [`spin_channel`](https://github.com/nomad-coe/nomad-simulations/blob/144826fb96ba339d97e94c13e60f0e7fedc1810a/src/nomad_simulations/schema_packages/properties/spectral_profile.py#L141-L146) 'can take values of 0 or 1', the [normalization factor](https://github.com/nomad-coe/nomad-simulations/blob/144826fb96ba339d97e94c13e60f0e7fedc1810a/src/nomad_simulations/schema_packages/properties/spectral_profile.py#L347-L353) is halved whenever it is set ('If spin_channel is set, assume spin-polarized system'), and [`extract_spin_polarized_property`](https://github.com/nomad-coe/nomad-simulations/blob/144826fb96ba339d97e94c13e60f0e7fedc1810a/src/nomad_simulations/schema_packages/outputs.py#L213-L233) classifies any section with a set value as spin-polarized. A restricted DOS tagged `spin_channel=0` therefore gets the wrong normalization. The old `get_total_dos` left it unset; `None` is dropped during conversion, so:
```suggestion
                spin=spin if n_spin == 2 else None,
```

src/nomad_simulation_parsers/parsers/vasp/doscar_parser.py
ladinesa: done

src/nomad_simulation_parsers/parsers/vasp/doscar_parser.py
ladinesa: The file parser expects a DOSCAR format but tries to returns whatever block it parses. It is up to the mapping parser if this make sense or not, if not should really not return any dos data. I have to correct the test as it is not confirming with the DOSCAR format.

src/nomad_simulation_parsers/schema_packages/vasp.py
ladinesa: remove l500

src/nomad_simulation_parsers/schema_packages/vasp.py
ladinesa: set to append

src/nomad_simulation_parsers/parsers/vasp/doscar_parser.py
ladinesa: done

src/nomad_simulation_parsers/parsers/vasp/doscar_parser.py
ladinesa: I think this is then a problem with the normalizer. But I will set spin=None for nsp here.

ndaelman-hu: I opened an issue for the schema now, as explained in https://github.com/FAIRmat-NFDI/nomad-parser-plugins-simulation/pull/207#discussion_r3750435525

src/nomad_simulation_parsers/schema_packages/vasp.py
ndaelman-hu: Fine as-is — I checked: `merge@last` currently behaves like `append` across passes anyway.

src/nomad_simulation_parsers/parsers/vasp/doscar_parser.py
ndaelman-hu: Fair enough. Then pls also drop the minimal-file special-casing in `DOSCARFileParser` (`n_spin = max(1, ...)`) — sub-spec files currently pass it only to vanish in the swallowed `IndexError` here.

ladinesa: done

src/nomad_simulation_parsers/parsers/vasp/doscar_parser.py
```diff
@@ -0,0 +1,172 @@
...
+        return self.data_object._results or {}
+
+    def get_dos(
+        self, total: np.ndarray, projected: list[dict[str, Any]]
+    ) -> list[dict[str, Any]]:
+        n_spin = max(1, len(total) // 2)
+        dos = []
+        for spin in range(n_spin):
+            dct = dict(
+                dos=total[spin],
```
ndaelman-hu: Following up on the `np.abs` note from the review body, now with a concrete failure: `ElectronicDensityOfStates.value` is `positive_float()` in the schema and validates on assignment, and the mapping parser swallows the `ValueError` — so a DOSCAR containing any negative DOS value silently loses its total DOS. Checked with `AgAc_relax` and a single negated entry: `value` ends up unset while `projected_dos` survives. Negative dips are routine with Methfessel-Paxton smearing, so real uploads will hit this. It also leaves the two DOS paths inconsistent: the vasprun path (`xml_parser.py`) still stores `np.abs`.

Restoring `np.abs` is not quite right either — it turns the smearing's negative lobes into spurious peaks. Suggest clipping instead, as a workaround until FAIRmat-NFDI/nomad-simulations#461 settles how the schema should handle these:
```suggestion
                dos=np.maximum(total[spin], 0),
```
For coverage: NOMAD entry `08josg0flNXtN7ju_fEtcXsjHm8_` (Fe, `ISPIN=2`) has a real 5-column DOSCAR plus OUTCAR that would exercise the spin-polarized branch here; a negative dip in a synthetic block (like the suffix test) would pin the clipping.

ndaelman-hu (approved):
Approving: the consolidation is additive (the Lobster archives are byte-identical to `develop`, the OUTCAR path gains projected DOS) and the review points were addressed. Two items remain, neither blocking:
- Inline below: negative DOS values currently vanish silently; interim clip suggested, schema-side handling tracked in FAIRmat-NFDI/nomad-simulations#461.
- The `DOSCARFileParser` minimal-file special-casing ask in the earlier thread stands.

src/nomad_simulation_parsers/parsers/vasp/doscar_parser.py
ladinesa: ~~I think this should be handled by the normalizer~~, anyhow, I took your suggestion and added a warning. Ok this is a datatype so normalizer cannot help, but a default converter for this datatype should be implemented.

ladinesa: Regarding the minimal file case, I added an immediate return when only a single column is parsed, should I assert that an even number of columns is parsed?

ndaelman-hu: Yes, evenness is the right check — each spin channel contributes a dos + integrated column pair. Pls log a warning on that path though, so malformed files show up in the logs.

ladinesa: done

